PR #41, opened 2020-03-10
Allow infrastructure if state is empty + worker deletion if credentials changed

### LINKED ISSUE #12: Cannot delete infrastructure when credentials data keys a re missing in secret
From https://github.com/gardener/gardener-extensions/issues/577

> If the account secret does not contain a service account json, the cluster can for sure not be created.
But when trying to delete such a cluster this fails because of the same reason:
> ```
> Waiting until shoot infrastructure has been destroyed
> Last Error
> task "Waiting until shoot infrastructure has been destroyed" failed: Failed to delete infrastructure: Error deleting infrastructure: secret shoot--berlin--rg-kyma/cloudprovider doesn't have a service account json
> ```

> It is the same also for the other providers. This is not something specific to gcp.

### LINKED ISSUE #13: Update credentials during Worker deletion
From https://github.com/gardener/gardener-extensions/issues/523

> Steps to reproduce:
> 1. Create a Shoot with valid cloud provider credentials `my-secret`.
> 1. Ensure that the Shoot is successfully created.
> 1. Invalidate the `my-secret` credentials.
> 1. Delete the Shoot.
> 1. Update `my-secret` credentials with valid ones.
> 1. Ensure that the Shoot deletion fails waiting the Worker to be deleted.

>  Currently we do no sync the cloudprovider credentials in the `<Provider>MachineClass` during Worker deletion. Hence machine-controller-manager fails to delete the machines because the credentials are the invalid ones.

rfranzke:
This PR allows the deletion of the `Infrastructure` resource in case the state is empty and the secret does not contain the required credentials (they would not be needed anyways in this case).

Also, due to revendoring g/gext, the generic Worker actuator will make sure that the machine class secrets are updated pre deletion.

Fixes #12
Fixes #13

**Release note**:

```improvement user
The `Infrastructure` deletion does now succeed in case it never created anything but the cloud provider secret does not contain the correct credentials.
```
```improvement user
The `Worker` deletion does now succeed in case the cloud provider credentials were changed.
```


minchaow:
Hi @rfranzke, I am not sure if the PR would completely fix the issue #12. 
The deleting process seems not be blocked by the empty terraform state, more likely be blocked by the code https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/pkg/controller/infrastructure/actuator.go#L439-L442. Because the credentials are empty or incorrect, the `NewSLBClient` returns one error, which fails the deleting process.

rfranzke:
@minchaow The #12 issue is about creating an `Infrastructure` with invalid credentials and then trying to delete it - it will never succeed. This problem is fixed with this PR - if the state is empty then it doesn't matter whether the credentials are invalid or not - they are not needed because nothing was created in the first place.

The [`cleanupServiceLoadBalancers`](https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/pkg/controller/infrastructure/actuator.go#L433) function is only called later (after the check whether the state is empty or not).

Am I missing something / do you have another issue in mind?

minchaow:
@rfranzke  thanks for explanation, I missed this part of code:

> stateIsEmpty := tf.IsStateEmpty()
> if stateIsEmpty {
>   a.logger.Info("exiting early as infrastructure state is empty - nothing to do")
>   return tf.CleanupConfiguration(ctx)
> }

It is ok now, no other concern from my side.

PR #49, opened 2020-03-20
Add validating webhook for shoots

### LINKED ISSUE #15: Provider-specific webhooks in Garden cluster
From https://github.com/gardener/gardener-extensions/issues/407

> With the new `core.gardener.cloud/v1alpha1.Shoot` API Gardener does no longer understand the provider-specifics, e.g., the infrastructure config, control plane config, worker config, etc.
This allows end-users to harm themselves and create invalid `Shoot` resources the Garden cluster. Errors will only become present during reconciliation part creation of the resource.

> Also, it's not possible to default any of the provider specific sections. Hence, we could also think about mutating webhooks in the future.

> As we are using the controller-runtime maintained by the Kubernetes SIGs it should be relatively easy to implement these webhooks as the library abstracts already most of the things.

> We should have a separate, dedicated binary incorporating the webhooks for each provider, and a separate Helm chart for the deployment in the Garden cluster.

> Similarly, the networking and OS extensions could have such webhooks as well to check on the `providerConfig` for the networking and operating system config.

> Part of https://github.com/gardener/gardener/issues/308

minchaow:
This PR adds a validating webhook that is responsible for checking resources of the group core.gardener.cloud (at the moment only shoot resources).

With [GEP-4](https://github.com/gardener/gardener/blob/master/docs/proposals/04-new-core-gardener-cloud-apis.md) cloud specific information was externalized and as a consequence cannot be validated by the Gardener anymore. On the other hand, it should be prevented that invalid shoot objects are created or updated.
Fixes #15

**Release note**:

```improvement operator
The `gardener-extension-provider-alicloud` now offers a validating webhook which checks shoot resources of type `alicloud` in the `core.gardener.cloud` group. Operators should register this webhook in the Garden cluster to further prevent invalid modifications on `alicloud` shoots.
```


.ci/pipeline_definitions
```diff
@@ -12,6 +12,11 @@ gardener-extension-provider-alicloud:
             image: 'eu.gcr.io/gardener-project/gardener/extensions/provider-alicloud'
             dockerfile: 'Dockerfile'
             target: gardener-extension-provider-alicloud
+          gardener-extension-provider-alicloud:
```
rfranzke: This should be `ardener-extension-validator-alicloud`

minchaow: good catch! Thanks a lot. I will fix it immediately.

PR #69, opened 2020-04-27
fix cannot delete infra when credentials are invalid

minchaow:
This PR is to fix issue that cannot delete infra when credentials are invalid.
Fixes #
#12 
Because need to Unmarshal `tf.state` into json format, so in this change I defined `tfState` and `tfStateResource` structure. I noticed that repo gardener-extension already contained `tf.state` related structure definition like `terraformStateV4`, so is it better to put these definition together, or separately because currently only Alicloud need to check the resources in `tf.state`?

**Release note**:

```improvement operator
Fix issue #12 cannot delete infra when credentials are invalid on alicloud
```


pkg/controller/infrastructure/actuator.go
```diff
@@ -484,6 +484,15 @@ func (a *actuator) Delete(ctx context.Context, infra *extensionsv1alpha1.Infrast
 		return tf.CleanupConfiguration(ctx)
 	}
 
+	resIsEmpty, err := common.IsResourceEmpty(tf)
```
rfranzke, line 487: I would combine lines 481-485 in this new `IsResourceEmpty` function somehow.

pkg/controller/common/terraform.go
```diff
@@ -61,3 +76,27 @@ func TerraformVariablesEnvironmentFromCredentials(credentials *alicloud.Credenti
...
+		return false, err
+	}
+
+	var state tfState
+	if err := json.Unmarshal(stateConfigMap, &state); err != nil {
+		return false, err
+	}
+
+	for _, res := range state.Resources {
+		if res.Provider == TerraformProvider && len(res.Instances) > 0 {
```
rfranzke, line 96: Is the provider check needed? Obviously, you can assume that the Alicloud extension only schedules Alicloud Terraform jobs, or?

minchaow: the provider check is necessary, because currently we have one `null provider` for workaround, see https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/charts/internal/alicloud-infra/templates/main.tf#L102-L106.

it always will generate not empty instances in `tf.state` like:
```
{
    "mode": "managed",
    "type": "null_resource",
    "name": "outputs",
    "provider": "provider.null",
    "instances": [
        {
            "schema_version": 0,
            "attributes": {
                "id": "4543855011011788166",
                "triggers": {
                    "recompute": "outputs"
                }
            }
        }
    ]
}
```

pkg/controller/infrastructure/actuator.go
minchaow: good suggestion, I will combine these lines together

pkg/controller/common/terraform.go
rfranzke: 👍 Makes sense, thanks for the explanation!

PR #77, opened 2020-05-12
enable specified eips on snat

minchaow:
This PR is to enable specified eips on snat of natgateway. Users can enable this feature by the following configuration:
```yaml
apiVersion: alicloud.provider.extensions.gardener.cloud/v1alpha1
kind: InfrastructureConfig
networks:
    vpc:
        cidr: 10.242.0.0/16
    zones:
        - name: cn-shanghai-e
           workers: 10.242.0.0/19
           natGateway:
               ipAddresses:
               - 10.11.12.13
```
Currently alicloud only support one specified eip, because if users want to bind multiply eips on one vsw, they have to enable `Shared Bandwidth Package` service.

If users enable this feature on already existed clusters, the origin auto-generated eip would be destroyed and bind the specified eip on snat.

**DONOT merge this PR because**
1. It is blocked by this terraform issue of alicloud provider:
https://github.com/terraform-providers/terraform-provider-alicloud/issues/2399

2. It needs to update `terraform` version to `0.12.20`:
https://github.com/gardener/terraformer/pull/39

Fixes #
#55
None

**Release note**:

```improvement operator
Alicloud now supports the feature of specifying eips on SNAT of NatGateway. If enable this feature on already existed clusters, origin auto-generated eip would be destroyed.
```


pkg/apis/alicloud/v1alpha1/types_infrastructure.go
```diff
@@ -95,6 +95,13 @@ type Zone struct {
 	Worker string `json:"worker"`
 	// Workers specifies the worker CIDR to use.
 	Workers string `json:"workers"`
+	// NatGatewayConfig specifies the eips to bind
```
rfranzke, line 98:
```suggestion
	// NatGatewayConfig specifies the eips to bind
	// +optional
```

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
```diff
@@ -95,6 +95,13 @@ type Zone struct {
...
+	NatGateway NatGatewayConfig `json:"natGateway,omitempty"`
```
rfranzke, line 99:
```suggestion
	NatGateway *NatGatewayConfig `json:"natGateway,omitempty"`
```

pkg/apis/alicloud/types_infrastructure.go
```diff
@@ -94,6 +94,13 @@ type Zone struct {
 	Worker string
 	// Workers specifies the worker CIDR to use.
 	Workers string
+	// NatGatewayConfig specifies the eips to bind
+	NatGateway NatGatewayConfig
```
rfranzke, line 98:
```suggestion
	NatGateway *NatGatewayConfig
```

rfranzke:
Ref: https://github.com/gardener/gardener-extension-provider-aws/pull/96

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
minchaow: done

pkg/apis/alicloud/types_infrastructure.go
minchaow: done

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
minchaow: done

charts/internal/alicloud-infra/templates/main.tf
```diff
@@ -33,6 +33,25 @@ resource "alicloud_vswitch" "vsw_z{{ $index }}" {
   availability_zone = "{{ required "zone.name is required" $zone.name }}"
 }
 
+{{ if $zone.natGateway }}
+{{ if $zone.natGateway.ipAddresses -}}
+// Use specified EIP
+data "alicloud_eip" "eip_natgw_ds{{ $index }}" {
+  ip_addresses = "[{{ index $zone.natGateway.ipAddresses 0 }}]"
```
rfranzke, line 40: Why is
```yaml
  natGateway:
    ipAddresses:
    - 10.11.12.13
```

a list when you only access index `0`? Why not looping here? If it's not possible, I suggest to change the `values.yaml` structure from `list` to `string`.

pkg/apis/alicloud/types_infrastructure.go
```diff
@@ -94,6 +94,14 @@ type Zone struct {
...
 	// Workers specifies the worker CIDR to use.
 	Workers string
+	// NatGatewayConfig specifies the eips to bind
+	// +optional
+	NatGateway *NatGatewayConfig
+}
+
+// NatGatewayConfig contains specified eips to bind nat gateway
+type NatGatewayConfig struct {
+	IPAddresses []string
```
rfranzke, line 104: Doc string missing

Also, see above comment: Why taking a list here if you only use the first one?

pkg/apis/alicloud/types_infrastructure.go
```diff
@@ -94,6 +94,14 @@ type Zone struct {
 	Worker string
 	// Workers specifies the worker CIDR to use.
 	Workers string
+	// NatGatewayConfig specifies the eips to bind
+	// +optional
```
rfranzke, line 98:
```suggestion
```

not needed in internal API

pkg/apis/alicloud/types_infrastructure.go
```diff
@@ -94,6 +94,14 @@ type Zone struct {
 	Worker string
 	// Workers specifies the worker CIDR to use.
 	Workers string
+	// NatGatewayConfig specifies the eips to bind
```
rfranzke, line 97:
```suggestion
	// NatGatewayConfig specifies configuration for the NAT gateway in this zone.
```

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
```diff
@@ -95,6 +95,14 @@ type Zone struct {
 	Worker string `json:"worker"`
 	// Workers specifies the worker CIDR to use.
 	Workers string `json:"workers"`
+	// NatGatewayConfig specifies the eips to bind
```
rfranzke, line 98:
```suggestion
	// NatGatewayConfig specifies configuration for the NAT gateway in this zone.
```

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
```diff
@@ -95,6 +95,14 @@ type Zone struct {
...
+	// +optional
+	NatGateway *NatGatewayConfig `json:"natGateway,omitempty"`
+}
+
+// NatGatewayConfig contains specified eips to bind nat gateway
+type NatGatewayConfig struct {
+	IPAddresses []string `json:"ipAddresses,omitempty"`
```
rfranzke, line 105:
```suggestion
  // <doc string here, please>
  // +optional
	IPAddresses []string `json:"ipAddresses,omitempty"`
```

+ see above comment wrt `[]string` vs `string`.

pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -66,6 +67,8 @@ func ValidateInfrastructureConfig(infra *apisalicloud.InfrastructureConfig, node
 			allErrs = append(allErrs, cidrvalidation.ValidateCIDRIsCanonical(workerPath, zone.Workers)...)
 			workerCIDRs = append(workerCIDRs, cidrvalidation.NewCIDR(zone.Workers, workerPath))
 		}
+
+		allErrs = append(allErrs, ValidateNatGatewayConfig(zone.NatGateway, networksPath.Child("zones").Index(i).Child("natGateway").Child("ipAddresses"))...)
```
rfranzke, line 71:
```suggestion
		allErrs = append(allErrs, ValidateNatGatewayConfig(zone.NatGateway, networksPath.Child("zones").Index(i).Child("natGateway"))...)
```

pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -97,8 +100,50 @@ func ValidateInfrastructureConfig(infra *apisalicloud.InfrastructureConfig, node
...
+	}
+
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks.VPC, oldConfig.Networks.VPC, field.NewPath("networks").Child("vpc"))...)
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks.Zones, oldConfig.Networks.Zones, field.NewPath("networks").Child("zones"))...)
+	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newZonesWithoutNatGatewayConfig, oldZonesWithoutNatGatewayConfig, field.NewPath("networks").Child("zones"))...)
+
+	return allErrs
+}
+
+func ValidateNatGatewayConfig(natGateway *apisalicloud.NatGatewayConfig, fldPath *field.Path) field.ErrorList {
```
rfranzke, line 136: doc string

pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -97,8 +100,50 @@ func ValidateInfrastructureConfig(infra *apisalicloud.InfrastructureConfig, node
...
+	allErrs := field.ErrorList{}
+
+	if natGateway != nil && natGateway.IPAddresses != nil {
+		if len(natGateway.IPAddresses) > 1 {
+			allErrs = append(allErrs, field.Invalid(fldPath, natGateway.IPAddresses, "currently can only specify one eip"))
+		}
+		if ipAddr := net.ParseIP(natGateway.IPAddresses[0]); ipAddr == nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(0), natGateway.IPAddresses[0], "specified eip is not valid"))
+		}
```
rfranzke, line 145: I'd still loop here - then you don't have to adapt the code when the above restriction (only 1 IP) is lifted.

pkg/controller/infrastructure/terraform.go
```diff
@@ -64,12 +64,25 @@ func (terraformOps) ComputeChartValues(
 		if workersCIDR == "" {
 			workersCIDR = zone.Worker
 		}
-		zones = append(zones, map[string]interface{}{
+
+		z := map[string]interface{}{
```
rfranzke, line 68:
```suggestion
		zoneConfig := map[string]interface{}{
```

pkg/controller/infrastructure/terraform.go
```diff
@@ -64,12 +64,25 @@ func (terraformOps) ComputeChartValues(
...
 			},
-		})
+		}
+
+		if zone.NatGateway != nil {
+			ipAddresses := make([]string, 0, len(zone.NatGateway.IPAddresses))
+			ipAddresses = append(ipAddresses, zone.NatGateway.IPAddresses...)
+
+			natGateway := make(map[string]interface{})
+			natGateway["ipAddresses"] = ipAddresses
```
rfranzke, line 80:
```suggestion
			natGateway := make(map[string]interface{})
			natGateway["ipAddresses"] = zone.NatGateway.IPAddresses
```

pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -97,8 +100,50 @@ func ValidateInfrastructureConfig(infra *apisalicloud.InfrastructureConfig, node
...
+			Worker:  zone.Worker,
+			Workers: zone.Workers,
+		})
+
+		allErrs = append(allErrs, ValidateNatGatewayConfig(zone.NatGateway, networksPath.Child("zones").Index(i).Child("natGateway").Child("ipAddresses"))...)
+	}
+
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks.VPC, oldConfig.Networks.VPC, field.NewPath("networks").Child("vpc"))...)
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks.Zones, oldConfig.Networks.Zones, field.NewPath("networks").Child("zones"))...)
+	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newZonesWithoutNatGatewayConfig, oldZonesWithoutNatGatewayConfig, field.NewPath("networks").Child("zones"))...)
```
rfranzke, line 131: TBH, I don't really understand this code. What is happening here? Is there a unit test for it?

charts/internal/alicloud-infra/templates/main.tf
minchaow: Thanks for your detailed review on this change.  Currently, alicloud only supports binding multi eips on one vswitch only when `shared bandwidth package` service has been enabled, so customers cannot bind multi eips without `shared bandwidth package` service. The reason why I use `list` is for the consideration of future change adaptation if alicloud allow binding multi eips without `shared bandwidth package` service in the future. In that case, we don't need to change the data structure, and do some change on `main.tf` is ok.

Of course, currently because only one eip can be bound without `shared bandwidth package` service, we cannot use `string` rather than `list`. 

Which one is preferable?

rfranzke: OK, thanks for the clarification. If you already know that it'll be possible in the future to bind multiple EIPs to it then it makes sense to keep the list. If that's just a hope/wish without concrete plans from Alicloud's side then I would use a `string`. :)

pkg/apis/alicloud/validation/infrastructure.go
minchaow: The code is to allow user update the field `natGateway` under the `zones`, and they have the follow structure.
```
networks:
  vpc:
    cidr: 10.250.0.0/16
  zones:
  - name: cn-shanghai-e
    worker: 10.250.0.0/19
    natGateway:
      ipAddresses: 
```
Because of this code `allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks.Zones, oldConfig.Networks.Zones, field.NewPath("networks").Child("zones"))...)`, it will forbid the update of all fields under `zones`. In order to make `natGateway` could be updated and other fields not allowed, I generate two new `Zones` structure without `natGateway` info to validate if there is some change or not on other fields. 

If I have some misunderstanding of the code logic, please correct me. :)

charts/internal/alicloud-infra/templates/main.tf
```diff
@@ -33,6 +33,25 @@ resource "alicloud_vswitch" "vsw_z{{ $index }}" {
   availability_zone = "{{ required "zone.name is required" $zone.name }}"
 }
 
+{{ if $zone.natGateway }}
+{{ if $zone.natGateway.ipAddresses -}}
```
jia-jerry, line 37: What if ipAddress is empty? Shall we just conbine the check with `and`?

pkg/apis/alicloud/validation/infrastructure.go
rfranzke: Ah, interesting 😄 I see, understood now. Maybe a comment would be helpful for readers.

pkg/apis/alicloud/validation/infrastructure.go
rfranzke: ping, still missing

pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -97,8 +100,48 @@ func ValidateInfrastructureConfig(infra *apisalicloud.InfrastructureConfig, node
...
+
+	return allErrs
+}
+
+func ValidateNatGatewayConfig(natGateway *apisalicloud.NatGatewayConfig, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if natGateway != nil && natGateway.IPAddress != nil {
+		ipAddr := string(*natGateway.IPAddress)
+		if ipAddrParsed := net.ParseIP(ipAddr); ipAddrParsed == nil {
```
rfranzke, line 141: Hm, out of curiosity - why not let people specify the EIP ID instead of the IP itself?

pkg/controller/infrastructure/terraform.go
```diff
@@ -64,12 +64,21 @@ func (terraformOps) ComputeChartValues(
...
 			"cidr": map[string]interface{}{
 				"workers": string(workersCIDR),
 			},
-		})
+		}
+
+		if zone.NatGateway != nil && zone.NatGateway.IPAddress != nil {
+			natGateway := make(map[string]interface{})
+			natGateway["ipAddress"] = string(*zone.NatGateway.IPAddress)
+			zoneConfig["natGateway"] = natGateway
```
rfranzke, line 78: The `string()` should not be needed. Also, I'd write it this way for better readability (though, that's just my personal preference ;-)):
```suggestion
			zoneConfig["natGateway"] = map[string]interface{}{
					"ipAddress": *zone.NatGateway.IPAddress,
			}
```

pkg/apis/alicloud/validation/infrastructure.go
minchaow: Hm, I think `IP address` is more readable and understandable compared with `EIP ID`, so let people able to configure the ip address. Also it would be more convenient for people to retrieve the ip address. (directly from gardener dashboard, not necessary to login Alicloud console).

Of course, if let people specify `EIP ID`, our code can directly bind this value on `vswitch` in the TF scripts, no need to obtain it from `data` reosource.

rfranzke: Yeah, just saying that it's inconsistent now, as people would need to specify a VPC _ID_ instead of a _name_ when they want to re-use an existing VPC. But it's your decision, just got curious. :)

minchaow: ah...good point! I missed that it would be inconsistent with re-using an existing VPC feature. I will modify my code, and make the behavior still consistent. :)

rfranzke: Maybe you can allow both :)

minchaow:
After a discussion with @jia-jerry, we finally decide to only support users to specify `eip id` to bind their own `eip` on `snat`. Mainly for the following two reasons:
1. to be consistent with feature of configurable vpc, which only supporting to specify `vpc id`.
2. make the solution easier and more maintainable in the future compared with allow both `eip address` and `eip id` specification solution.

pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -97,8 +98,48 @@ func ValidateInfrastructureConfig(infra *apisalicloud.InfrastructureConfig, node
...
+
+	return allErrs
+}
+
+// ValidateNatGatewayConfig validates a NatGatewayConfig object.
+func ValidateNatGatewayConfig(natGateway *apisalicloud.NatGatewayConfig, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if natGateway != nil {
+		if natGateway.EIPAllocationID == nil {
```
jia-jerry, line 139: if EIPAllocationID == "", will terraform run with error?

pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -97,8 +98,48 @@ func ValidateInfrastructureConfig(infra *apisalicloud.InfrastructureConfig, node
...
+	}
+
+	for i, zone := range oldConfig.Networks.Zones {
+		oldZonesWithoutNatGatewayConfig = append(oldZonesWithoutNatGatewayConfig, apisalicloud.Zone{
+			Name:    zone.Name,
+			Worker:  zone.Worker,
+			Workers: zone.Workers,
+		})
+
+		allErrs = append(allErrs, ValidateNatGatewayConfig(zone.NatGateway, networksPath.Child("zones").Index(i).Child("natGateway"))...)
```
jia-jerry, line 125: Do we need to validate old config?

minchaow: I don't think we need to validate old config, because after we update nat gateway config, we will use the new config to generate `TF` scripts, never to use the old scripts again. And `TF` will help us the reconcile the relationship between the new and old config.

pkg/apis/alicloud/validation/infrastructure.go
minchaow: if `EIPAllocationID` equals to "", it will generate the following TF scripts:
```
data "alicloud_eips" "eip_natgw_ds{{ $index }}" {
  ids = [""]
}
```
so the 'TF' will run with error, because "" is treated as invalidate value of `eip` id.

pkg/apis/alicloud/validation/infrastructure.go
minchaow: already removed validation of old config

PR #79, opened 2020-05-14
Use Terraform resource timeouts and workqueue backoff

rfranzke:
With this PR we add some timeouts for those Terraform resources that are supporting it. Also, we will no longer requeue with the constant `30s` on errors but leverage the exponential backoff functionality of the underlying workqueue.

Part of gardener/gardener#2253

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/charts/internal/alicloud-infra/templates/main.tf
+++ b/charts/internal/alicloud-infra/templates/main.tf
@@ -14,7 +14,13 @@ resource "alicloud_key_pair" "publickey" {
 resource "alicloud_vpc" "vpc" {
   name       = "{{ required "clusterName is required" .Values.clusterName }}-vpc"
   cidr_block = "{{ required "vpc.cidr is required" .Values.vpc.cidr }}"
+
+  timeouts {
+    create = "5m"
+    delete = "5m"
+  }
 }
+
 resource "alicloud_nat_gateway" "nat_gateway" {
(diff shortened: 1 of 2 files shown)


rfranzke:
/area/quality
/kind/enhancement
/needs/review
/priority/normal
/size/xs

charts/internal/alicloud-infra/templates/main.tf
```diff
@@ -14,7 +14,13 @@ resource "alicloud_key_pair" "publickey" {
 resource "alicloud_vpc" "vpc" {
   name       = "{{ required "clusterName is required" .Values.clusterName }}-vpc"
   cidr_block = "{{ required "vpc.cidr is required" .Values.vpc.cidr }}"
+
+  timeouts {
```
minchaow, line 18: why don't have timeouts setting for `update`?

rfranzke: It's not supported: https://www.terraform.io/docs/providers/alicloud/r/vpc.html#timeouts

rfranzke:
> I am also wondering why only vpc and nat_gateway resources have timeouts setting? how about other resources?

The timeouts are resource specific and not all resources support it, see https://www.terraform.io/docs/configuration/resources.html#operation-timeouts

minchaow:
@rfranzke: Thanks for your explanation. 
/lgtm

PR #89, opened 2020-05-29
support injecting PROVIDER_ID in system environment

jia-jerry:
/area control-plane
/kind enhancement
/priority blocker
/platform alicloud

Inject environment PROVIDER_ID before kubelet is started
Fixes https://github.com/gardenlinux/gardenlinux/issues/67

**Release note**:

```improvement operator
Gardenlinux is upported on Alicloud now.
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -113,6 +113,24 @@ func (e *ensurer) EnsureKubeletServiceUnitOptions(ctx context.Context, ectx gene
 		command = ensureKubeletCommandLineArgs(command)
 		opt.Value = extensionswebhook.SerializeCommandLine(command, 1, " \\\n    ")
 	}
+
+	/*
+	 * # Set environment PROVIDER_ID to /var/lib/kubelet/extra_args
+	 *
+	 * grep -sq PROVIDER_ID /var/lib/kubelet/extra_args
+	 * if [ $? -ne 0 ]; then
+	 *   META_EP=http://100.100.100.200/latest/meta-data
+	 *   PROVIDER_ID=`wget -qO- $META_EP/region-id $META_EP/region-id`.`wget -qO- $META_EP/region-id $META_EP/instance-id`
+	 *   echo PROVIDER_ID=$PROVIDER_ID >> /var/lib/kubelet/extra_args
+	 *   echo PROVIDER_ID=$PROVIDER_ID has been written to /var/lib/kubelet/extra_args
+	 * fi
+	 */
+	new = extensionswebhook.EnsureUnitOption(new, &unit.UnitOption{
+		Section: "Service",
+		Name:    "ExecStartPre",
+		//This doesn't work: /bin/sh -c "$(echo  Z3JlcCAtc3EgUFJPVklERVJfSUQgL2V0Yy9lbnZpcm9ubWVudAppZiBbICQ.... | base64 -d)"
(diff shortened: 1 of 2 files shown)


pkg/webhook/controlplane/ensurer.go
```diff
@@ -113,6 +113,24 @@ func (e *ensurer) EnsureKubeletServiceUnitOptions(ctx context.Context, ectx gene
...
+
+	/*
+	 * # Set system environment PROVIDER_ID
+	 *
+	 * grep -sq PROVIDER_ID /etc/environment
+	 * if [ $? -ne 0 ]; then
+	 * 	META_EP=http://100.100.100.200/latest/meta-data
+	 * 	PROVIDER_ID=`curl -s $META_EP/region-id`.`curl -s $META_EP/instance-id`
+	 * 	echo PROVIDER_ID=$PROVIDER_ID >> /etc/environment
+	 * 	echo PROVIDER_ID=$PROVIDER_ID has been written to /etc/environment
```
vpnachev, line 125: Instead of touching the `/etc/environment` file, you can try to set the variable directly in the service or as a dropin.
You could check these links:
- https://coreos.com/os/docs/latest/using-environment-variables-in-systemd-units.html
- https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html
- https://serverfault.com/questions/413397/how-to-set-environment-variable-in-systemd-service

jia-jerry: `Environment` is not static. It is composed @Virtual Machine. I can't think out of way of set `Environment` variable. As about drop in files, I didn't find a place to inject it. Maybe you could give me some hint.

jia-jerry: btw, I found in ExecStartPre, 
```bash
$(echo  Z3JlcCAtc3EgUFJPVklERVJfSUQgL2V0Yy9lbnZpcm9ubWVudAppZiBbICQ.... | base64 -d)
```
doesn't work

jia-jerry: `Environment` is not static. It is composed @Virtual Machine. I can't think out of way of set `Environment` variable. As about drop in files, I didn't find a place to inject it. Maybe you could give me some hint. BTW, inline statement like below can't be evaluated in service unit.
```bash
PROVIDER_ID=`curl -s $META_EP/region-id`.`curl -s $META_EP/instance-id`
```

vpnachev: I was thinking of a dropin configuration file that is setting the environment variables for the kubelet service
```
Environment=PROVIDER_ID=/path/to/provider_id
```

While an `ExecStartPre` is populating the file
```
...
ExecStartPre="/bin/sh -c 'echo $(curl -s $META_EP/region-id).$(curl -s $META_EP/instance-id) >/path/to/provider_id'"
...
```


But I have not tested it and it might not succeed to populate the file on time.
I am fine with the current solution - maybe use a dedicated file `/etc/enviroment.d/kubelet.conf` instead of writing in the global file.

jia-jerry: As about environment file, good suggestion. I chaged to `/var/lib/kubelet/extra_args`
Here is the ultimate kubelet.service
```
[Unit]
Description=kubelet daemon
Documentation=https://kubernetes.io/docs/admin/kubelet
After=docker.service
Wants=docker.socket rpc-statd.service

[Install]
WantedBy=multi-user.target

[Service]
Restart=always
RestartSec=5
EnvironmentFile=/etc/environment
EnvironmentFile=-/var/lib/kubelet/extra_args
ExecStartPre=/usr/bin/docker run --rm -v /opt/bin:/opt/bin:rw --entrypoint /bin/sh registry-vpc.cn-shanghai.aliyuncs.com/gardener-cn/k8s_gcr_io_hyperkube@sha256:4f547134137348fba5105f70874f2267b0c5a5ce33a924afd39142c14ee62af9 -c "cp /usr/local/bin/kubelet /opt/bin"
ExecStart=/opt/bin/kubelet --bootstrap-kubeconfig=/var/lib/kubelet/kubeconfig-bootstrap \
    --config=/var/lib/kubelet/config/kubelet \
    --cni-bin-dir=/opt/cni/bin/ \
    --cni-conf-dir=/etc/cni/net.d/ \
    --image-pull-progress-deadline=1m \
    --pod-infra-container-image=registry-vpc.cn-shanghai.aliyuncs.com/gardener-cn/google_containers_pause-amd64@sha256:fcaff905397ba63fd376d0c3019f1f1cb6e7506131389edbcb3d22719f1ae54d \
    --kubeconfig=/var/lib/kubelet/kubeconfig-real \
    --network-plugin=cni \
    --volume-plugin-dir=/var/lib/kubelet/volumeplugins \
    --v=2 \
    $KUBELET_EXTRA_ARGS \
    --provider-id=${PROVIDER_ID} \
    --cloud-provider=external \
    --enable-controller-attach-detach=true
ExecStartPre=/bin/sh -c "echo Z3JlcCAtc3EgUFJPVklERVJfSUQgL3Zhci9saWIva3ViZWxldC9leHRyYV9hcmdzCmlmIFsgJD8gLW5lIDAgXTsgdGhlbgpNRVRBX0VQPWh0dHA6Ly8xMDAuMTAwLjEwMC4yMDAvbGF0ZXN0L21ldGEtZGF0YQpQUk9WSURFUl9JRD1gY3VybCAtcyAkTUVUQV9FUC9yZWdpb24taWRgLmBjdXJsIC1zICRNRVRBX0VQL2luc3RhbmNlLWlkYAplY2hvIFBST1ZJREVSX0lEPSRQUk9WSURFUl9JRCA+PiAvdmFyL2xpYi9rdWJlbGV0L2V4dHJhX2FyZ3MKZWNobyBQUk9WSURFUl9JRD0kUFJPVklERVJfSUQgaGFzIGJlZW4gd3JpdHRlbiB0byAvdmFyL2xpYi9rdWJlbGV0L2V4dHJhX2FyZ3MKZmkK | base64 -d > /var/lib/kubelet/gardener-set-provider-id && chmod +x /var/lib/kubelet/gardener-set-provider-id && /var/lib/kubelet/gardener-set-provider-id"
```

PR #116, opened 2020-07-15
Enable whitelisting for machine image sharing

EmoinLanyu:

/area infrastructure
/area customized-image-sharing
/kind enhancement
/priority normal
/platform alicloud

**Background context:**
1. On Alicloud, users can upload and create their own machine images. However, if they are not explicitly shared, customized os images can only be used by the Alicloud account that creates them.
1. Gardener is now adopting Gardenlinux, which currently can only be used as a customized machine image on Alicloud.
1. Alicloud extension supported customized image sharing in [this PR](https://github.com/gardener-attic/gardener-extensions/pull/427). The feature was added so that Gardener can use customized os images on Alicloud.
1. In the past, when sharing an image from the image owner account to end-user's Alicloud account, we have to specify the account id of the receiver (example [here](https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/pkg/controller/infrastructure/actuator.go#L363)). In this scenario, when a customized image is shared to an end-user, the end-user can then see this image in his image list from the console or by using api (e.g., using aliyuncli `aliyun ecs DescribeImages`).
1. Now, Alicloud supports sharing of a customized image to all users. If we call the old way (share to specific account) *private sharing*, the new feature is like *public sharing*. Different from *private sharing*, when an image is "publicly" shared, the target end-user cannot see the shared image in his image list.

**Reason for enhancement:**
Gardener administrators can now decide to support either "private" image sharing or "public" image sharing. If they choose to use "public" image sharing, when a shoot infrastructure is reconciled, extension controller cannot find the image in user's cloud provider account (refer to **point 5** in **Background context**). And the controller currently also does not know whether a image is "publicly" shared or not, so it will then go ahead and share the image again to the end-user, thus this image will be "privately" shared again.

To prevent this from happening, we introduce a new feature called *image sharing whitelisting*, by adding below section (for more detailed configuration, please refer to the [guide](https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/docs/usage-as-operator.md)) in ControllerConfiguration, user can specify the images that will skip private image sharing:

```yaml
whitelistedImageIDs:
- <image_id1>
- <image_id2>
...
```

Fixes #

**Release note**:

```improvement operator
Gardener-extension-provider-alicloud now supports the configuration of customized image sharing whitelist. Extension controller will not share the whitelisted machine image to shoot's owner cloud provider account. The whitelist can be used when a customized machine image is already made publicly accessible manually.
```


EmoinLanyu:
It's strange that github template files are changed, I'm not quite sure how it happened (it was definitely not my manual change). If they should not be changed, please let me know and I'll manually change it back.

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -20,8 +20,7 @@ If multiple identifiers make sense you can also state the commands multiple time
 -->
 /area TODO
 /kind bug
-/priority normal
-/platform alicloud
+/priority normaln/platform alicloud
```
ialidzhikov, line 23: You could revert similar changes.

See https://github.com/gardener/gardener-extension-os-suse-chost/pull/22#discussion_r452649842.

EmoinLanyu: done

EmoinLanyu:
/invite @gardener/gardener-extension-provider-alicloud-maintainers

pkg/cmd/config.go
```diff
@@ -82,6 +82,12 @@ func (c *Config) ApplyMachineImageOwnerSecretRef(secretRef **corev1.SecretRefere
 	}
 }
 
+func (c *Config) ApplyWhitelistedImageIDs(whitelistedImageIDs **[]string) {
```
minchaow, line 85: why here is using one pointer of pointer of one slice?

pkg/controller/infrastructure/actuator.go
```diff
@@ -79,6 +80,7 @@ func NewActuatorWithDeps(
 	chartRendererFactory extensionschartrenderer.Factory,
 	terraformChartOps TerraformChartOps,
 	machineImageOwnerSecretRef *corev1.SecretReference,
+	whitelistedImageIDs *[]string,
```
minchaow, line 83: why using one pointer of slice?

EmoinLanyu: In case user does not provide a whitelist, `whitelistedImageIDs` would be nil. Then we can use `if whitelistedImageIDs == nil` to determine whether or not they are configured.

pkg/cmd/config.go
EmoinLanyu: `options.WhitelistedImageIDs` is `*[]string`, to assign value to it we need to get the address of it.

pkg/controller/infrastructure/actuator.go
minchaow: As I know, we can directly use `nil` to determine if the one slice is nil or not. So why not use `whitelistedImageIDs == nil` or `len(whitelistedImageIDs) == 0` to determine if the slice is nil or empty?

pkg/cmd/config.go
EmoinLanyu: will remove slice pointer

pkg/controller/infrastructure/actuator.go
EmoinLanyu: will remove slice pointer

PR #123, opened 2020-08-03
vendor g/g729a1cce46be

vpnachev:

/area dev-productivity
/kind task
/priority normal
/platform alicloud

vendor gardener/gardener@729a1cce46be

Fixes #

**Release note**:

```improvement operator
NONE
```


pkg/controller/infrastructure/actuator_test.go
```diff
@@ -54,6 +53,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 )
 
+const (
+	terraformMainTFFilename      = "main.tf"
+	terraformVariablesTFFilename = "variables.tf"
+	terraformTFVarsFilename      = "terraform.tfvars"
+)
```
rfranzke, line 60: You could use https://github.com/gardener/gardener/blob/master/extensions/pkg/terraformer/config.go#L32-L41

vpnachev: Good catch, I had to search for the value, not for the variable name.

PR #133, opened 2020-08-13
Recreate storage classes if update is failing

vpnachev:

/area ops-productivity robustness storage
/kind enhancement
/priority normal
/topology shoot
/platform alicloud

Recreate storage classes if update if failing

Ref https://github.com/gardener/gardener-resource-manager/pull/69

**Release note**:

```improvement operator
The `StorageClass`es  in the shoot cluster managed by Gardener are now re-created in case the update request failed due to changed immutable fields.
```

### DIFF
--- a/charts/internal/shoot-storageclasses/templates/storageclasses.yaml
+++ b/charts/internal/shoot-storageclasses/templates/storageclasses.yaml
@@ -5,6 +5,7 @@ metadata:
   name: default
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
+    resources.gardener.cloud/delete-on-invalid-update: "true"
 provisioner: diskplugin.csi.alibabacloud.com
 allowVolumeExpansion: true
 parameters:


rfranzke:
@minchaow some end-users delete our storage class, recreate it with the same name and different properties, which will lead to errors on outside requiring ops effort

jia-jerry:
> @minchaow some end-users delete our storage class, recreate it with the same name and different properties, which will lead to errors on outside requiring ops effort

Hmm, maybe we could also prevent user to delete managed resources by adding a finalizer.

rfranzke:
@jia-jerry I guess it doesn't help - then people would simply remove the finalizer and we are stuck again. Better go this "self-healing" way (that's how it's supposed to work (also for other system components))

jia-jerry:
> @jia-jerry I guess it doesn't help - then people would simply remove the finalizer and we are stuck again. Better go this "self-healing" way (that's how it's supposed to work (also for other system components))

PR #140, opened 2020-08-31
Configure tm bot to run infrastructure integration test

schrodit:

/kind test
/priority normal
/platform alicloud

Configures the testmachinery bot to run infrastructure integration test.
prerequisite for https://github.com/gardener/gardener-extension-provider-alicloud/pull/140

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/.ci/.testruns/default.yaml
+++ b/.ci/.testruns/default.yaml
@@ -1,0 +1,37 @@
+apiVersion: testmachinery.sapcloud.io/v1beta1
+kind: Testrun
+metadata:
+  generateName: tm-extension-alicloud-
+  namespace: default
+spec:
+
+  ttlSecondsAfterFinished: 172800 # 2 days
+
+  config:
+  - name: ACCESS_KEY_ID
+    type: env
+    valueFrom:
+      secretKeyRef:
+        name: shoot-operator-alicloud
+        key: accessKeyID
+  - name: ACCESS_KEY_SECRET
+    type: env
+    valueFrom:
+      secretKeyRef:
+        name: shoot-operator-alicloud
+        key: accessKeySecret
+
+  testflow:
+  - name: infrastructure-test
+    definition:
+      name: infrastructure-test
+      config:
+      - name: REGION
+        type: env
+        value: eu-central-1
+      - name: VPC_ID
+        type: env
+        valueFrom:
+          secretKeyRef:
+            name: alicloud-configuration
+            key: vpcID
--- a/.ci/tm-config.yaml
+++ b/.ci/tm-config.yaml
@@ -1,0 +1,2 @@
+test-single:
+  testrunPath: .ci/testruns/default.yaml


rfranzke (approved):
/lgtm

jia-jerry (changes requested):
@schrodit , Can we feed VPC-ID to the test run just like our e2e test. This VPC is attached with a pre-created NATGateway and it is shared with all testing shoots. The background is below. On AliCloud, NAT Gateway is charged daily and it is around 2 euros. Though AliCloud provided a beta version of NATGateway, which is charged hourly, but the API is different and we can't consume it so far.

@EmoinLanyu

schrodit:
@jia-jerry the test are executed in the same testmachinery cluster as the other tests. 
So it is possible, but I guess the mutating webhook has to be adjusted? cc @EmoinLanyu

EmoinLanyu:
> @jia-jerry the test are executed in the same testmachinery cluster as the other tests.
> So it is possible, but I guess the mutating webhook has to be adjusted? cc @EmoinLanyu

Hi @schrodit , mutating webhook will only mutate create-shoot Testruns, so this testrun will not be influenced, thus we don't need to adjust the webhook.

.ci/.testruns/default.yaml
```diff
@@ -0,0 +1,31 @@
...
+  ttlSecondsAfterFinished: 172800 # 2 days
+
+  config:
+  - name: ACCESS_KEY_ID
+    type: env
+    valueFrom:
+      secretKeyRef:
+        name: shoot-operator-alicloud
+        key: accessKeyID
+  - name: SECRET_ACCESS_KEY
```
minchaow, line 17: i think it should be `ACCESS_KEY_SECRET`

minchaow (changes requested):

schrodit:
@EmoinLanyu adding another env var for the VPC would be also possible.
But I thought that you also might want to add a subnet range so that this test does not interfere with the create shoot test?

I would also prefer to have it explicit in the testrun, if the ip ranges are no problem.

EmoinLanyu:
@schrodit We planned 31 subnets (total 32 subnets) in the VPC for testing, and 1 is left untouched. We can use the remaining subnet so that we don't interrupt other tests.

ialidzhikov:
> @schrodit , Can we feed VPC-ID to the test run just like our e2e test? This VPC is attached with a pre-created NATGateway and it is shared with all testing shoots. The background is below. On AliCloud, NAT Gateway is charged daily and it is around 2 euros. Though AliCloud provided a beta version of NATGateway, which is charged hourly, but the API is different and we can't consume it so far.

What is the reason to do not create the natGateway as part of the test?

schrodit:
> What is the reason to do not create the natGateway as part of the test?

The reasons are the costs of nat gateways in Alicloud. They are charged by day not minute or hour which had a massive affect on costs for alicloud.

But maybe this test is not affected as we do not run this testcase very often as we do with the default tests? 
@jia-jerry WDYT?

jia-jerry:
> > What is the reason to do not create the natGateway as part of the test?
> 
> The reasons are the costs of nat gateways in Alicloud. They are charged by day not minute or hour which had a massive affect on costs for alicloud.
> 
> But maybe this test is not affected as we do not run this testcase very often as we do with the default tests?
> @jia-jerry WDYT?

Ideally, we should also cover NAT Gateway  creation. I'm thinking the mechanism of infra-test triggering. Can we trigger the test with condition such as `main.tf` is changed in the PR?

schrodit:
> Ideally, we should also cover NAT Gateway creation. I'm thinking the mechanism of infra-test triggering. Can we trigger the test with condition such as main.tf is changed in the PR?

Currently this test is only triggered manually. So the maintainers decide when to trigger the test.

jia-jerry:
> > Ideally, we should also cover NAT Gateway creation. I'm thinking the mechanism of infra-test triggering. Can we trigger the test with condition such as main.tf is changed in the PR?
> 
> Currently this test is only triggered manually. So the maintainers decide when to trigger the test.

@minchaow  has already covered the 2 cases, creating NATGateway and reusing NATGateway. @schrodit, for reusing NATGateway case, I would like to use an existing VPC which e2e is using because it will save money :). So would you help configure VPC_ID env ?

jia-jerry (approved):

minchaow (approved):

PR #170, opened 2020-10-29
Forbid network cidr 100.64.0.0/10

### LINKED ISSUE #169: Prohibit user to create shoot within 100.64.0.0/10 cidr
**How to categorize this issue?**

/area networking
/kind enhancement
/priority normal
/platform alicloud

**What would you like to be added**:
Check service/pod/node network cidrs in shoot don't drop into 100.64.0.0/10 range.
**Why is this needed**:
Alicloud use 100.64.0.0/10 to provide VPC accessed services like DNS, Meta service and so one.

jia-jerry:

/area networking
/kind enhancement
/priority normal
/platform alicloud

Avoid user to input wrong cidrs for network settings
Fixes #169

**Release note**:

```improvement user
Users are forbidden to set cidrs for nodes/pods/services which overlap 100.64.0.0/10 because 100.64.0.0/10 is reverved by AliCloud
```

### DIFF
--- a/pkg/apis/alicloud/validation/shoot.go
+++ b/pkg/apis/alicloud/validation/shoot.go
@@ -19,21 +19,36 @@ import (
 	"regexp"
 
 	apisalicloud "github.com/gardener/gardener-extension-provider-alicloud/pkg/apis/alicloud"
-
 	"github.com/gardener/gardener/pkg/apis/core"
 	"github.com/gardener/gardener/pkg/apis/core/validation"
+	cidrvalidation "github.com/gardener/gardener/pkg/utils/validation/cidr"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/sets"
 	utilvalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
+// AliCloudVPCCidr is the IPV4 CIDR used within AliCloud.
+// For example: the meta service endpoint is 100.100.100.200.
+// This CIDR can be accessed by any machines which is running with AliCloud VPC.
+const AliCloudVPCCidr = "100.64.0.0/10"
+
 // ValidateNetworking validates the network settings of a Shoot.
 func ValidateNetworking(networking core.Networking, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if networking.Nodes == nil {
-		allErrs = append(allErrs, field.Required(fldPath.Child("nodes"), "a nodes CIDR must be provided for Alicloud shoots"))
+		allErrs = append(allErrs, field.Required(fldPath.Child("nodes"), "a nodes CIDR must be provided for AliCloud shoots"))
+	} else {
+		if cidrvalidation.NetworksIntersect(*networking.Nodes, AliCloudVPCCidr) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("nodes"), *networking.Nodes, "must not overlap with 100.64.0.0/10 because 10.64.0.0/10 is reserved by AliCloud"))
+		}
+	}
+	if cidrvalidation.NetworksIntersect(*networking.Pods, AliCloudVPCCidr) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("pods"), *networking.Pods, "must not overlap with 100.64.0.0/10 because 10.64.0.0/10 is reserved by AliCloud"))
+	}
+	if cidrvalidation.NetworksIntersect(*networking.Services, AliCloudVPCCidr) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("services"), *networking.Services, "must not overlap with 100.64.0.0/10 because 10.64.0.0/10 is reserved by AliCloud"))
 	}
 
 	return allErrs
--- a/pkg/apis/alicloud/validation/shoot.go
+++ b/pkg/apis/alicloud/validation/shoot.go
@@ -28,9 +28,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
-// AliCloudVPCCidr is the IPV4 CIDR used within AliCloud.
+// AliCloudVPCCidr is a IPV4 CIDR used within AliCloud.
 // For example: the meta service endpoint is 100.100.100.200.
-// This CIDR can be accessed by any machines which is running with AliCloud VPC.
+// This CIDR can be accessed by any machine which is running with AliCloud VPC.
 const AliCloudVPCCidr = "100.64.0.0/10"
 
 // ValidateNetworking validates the network settings of a Shoot.
--- a/pkg/apis/alicloud/validation/shoot.go
+++ b/pkg/apis/alicloud/validation/shoot.go
@@ -28,44 +28,45 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
-// AliCloudVPCCidr is a IPV4 CIDR used within AliCloud.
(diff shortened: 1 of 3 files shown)


rfranzke (commented):
What about shoots that already use this CIDR? If this change is introduced they cannot be updated/deleted anymore. 👀

jia-jerry:
> What about shoots that already use this CIDR? If this change is introduced they cannot be updated/deleted anymore. 👀

Good question. So far, no shoots are configured with this wrong cidr. I helped all shoots in ac-canary/ac-live to migrate the old original wrong default network setting(100.X.X.X/X) to correct one. If there is a chance that users specified a wrong one, I made a [tool](https://github.wdf.sap.corp/kubernetes-attic/shoot-network-hacker) to migrate. At least, I can be sure all productions are using correct CIDRs.

rfranzke:
I don't get this argumentation, we cannot look at specific/internal landscapes as the Alicloud extension is public/open-source, and there might be other landscapes that run it.

jia-jerry:
I know what your arguments because it is open sourced. I wantted to be arbitrary that shoots with network wrongly set can't be updated. The good thing is we provide an alarm and force users to take actions to avoid unexpected behaviors.

rfranzke:
Sorry, but I have to insist on that one. If we merge this PR then Alicloud shoots already using these CIDRs cannot be updated/deleted. The only option to merge this PR is to provide a document that explains what users have to do in that case (e.g., migrating the CIDRs with your script, or whatever is necessary to unblock the situation). If you want to provide this information, fine, otherwise, I suggest to adapt the implementation to ignore shoots that already have these CIDRs.

jia-jerry:
OK. I will add some logic to check the network only for creation phase.

jia-jerry:
One thing I don't understand, during update/creation, why should we check network originally? For update, node network is immutable(https://github.com/gardener/gardener/blob/master/pkg/apis/core/validation/shoot.go#L142) For creation, node network can't be nil(https://github.com/gardener/gardener/blob/master/pkg/apis/core/validation/shoot.go#L600).  I would suggest to check shoot network in ali-extention only for creation phase. WDYT?

rfranzke:
@jia-jerry 

> I would suggest to check shoot network in ali-extention only for creation phase and also remove nil check logic. WDYT? 

Once set, all CIDRs in `.spec.networking` (`pods`/`services`/`nodes`) are immutable: https://github.com/gardener/gardener/blob/master/pkg/apis/core/validation/shoot.go#L167-L175
So yes, it should be enough to only check the shoot networks in Alicloud extension when a new shoot is created because Gardener already ensures immutability on update. Good suggestion.

> BTW, which check comes first? Gardener adminsion vs Extension validation hook?

First Gardener static validation, then Gardener's own in-tree admission plugin, then external webhooks.

pkg/apis/alicloud/validation/shoot.go
```diff
@@ -19,21 +19,33 @@ import (
...
-	if networking.Nodes == nil {
-		allErrs = append(allErrs, field.Required(fldPath.Child("nodes"), "a nodes CIDR must be provided for Alicloud shoots"))
+	if networking.Nodes != nil && cidrvalidation.NetworksIntersect(*networking.Nodes, AliCloudVPCCidr) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("nodes"), *networking.Nodes, "must not overlap with 100.64.0.0/10 because 10.64.0.0/10 is reserved by AliCloud"))
+	}
+	if networking.Pods != nil && cidrvalidation.NetworksIntersect(*networking.Pods, AliCloudVPCCidr) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("pods"), *networking.Pods, "must not overlap with 100.64.0.0/10 because 10.64.0.0/10 is reserved by AliCloud"))
+	}
+	if networking.Services != nil && cidrvalidation.NetworksIntersect(*networking.Services, AliCloudVPCCidr) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("services"), *networking.Services, "must not overlap with 100.64.0.0/10 because 10.64.0.0/10 is reserved by AliCloud"))
```
rfranzke, line 48: All these fields (nodes/pods/services) are required for Alicloud. So you should enforce that these fields are not nil. Currently, you only check for intersection if they are not nil, but this would allow end-users to create a shoot without the nodes CIDR, for example.

pkg/apis/alicloud/validation/shoot.go
```diff
@@ -19,21 +19,33 @@ import (
...
+
+// ValidateNetworking validates the network settings of a Shoot during creation.
+// As the network settings are immutable, we don't need to do validation for update.
 func ValidateNetworking(networking core.Networking, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	if networking.Nodes == nil {
-		allErrs = append(allErrs, field.Required(fldPath.Child("nodes"), "a nodes CIDR must be provided for Alicloud shoots"))
+	if networking.Nodes != nil && cidrvalidation.NetworksIntersect(*networking.Nodes, AliCloudVPCCidr) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("nodes"), *networking.Nodes, "must not overlap with 100.64.0.0/10 because 10.64.0.0/10 is reserved by AliCloud"))
```
rfranzke, line 42: Don't copy the CIDR multiple times but use the `AliCloudVPCCidr` const (same for other two occurrences).

rfranzke (changes requested):

pkg/apis/alicloud/validation/shoot.go
jia-jerry: pods/services cidrs are ensured not nil in Gardener core service: https://github.com/gardener/gardener/blob/master/pkg/utils/validation/cidr/disjoint.go#L37
Regarding node cidrs, you are right. Still need to check nil.

rfranzke: Yes, but this validation will probably go away soon. Actually, that's a good example why any external code should not rely on what Gardener validates and what not - it is not subject to any API contract or reliable in any other way. This kind of invalidates what we discussed earlier, i.e., also on shoot updates the CIDR immutability should be checked.

jia-jerry: But from my perspective, I would like to check kubernetes network config (nodes/pods/service) in the core because these fields are exposed in Gardener core API. The extenstion validator only validates those specialties for cloudprovider specific.

jia-jerry: If such kind of validation will not happen in in-tree admission, shall we provide a generic check in common framwork of extension?

rfranzke: > because these fields are exposed in Gardener core API.

Yes, those are fields in the core, but it is infrastructure-specific whether they are required or not. For most hyperscalers they are required, but in private clouds or custom infrastructures not necessarily. Hence, the validation should happen here to ensure that ALL the fields needed by Alicloud extension are present.

> If such kind of validation will not happen in in-tree admission, shall we provide a generic check in common framwork of extension?

We could think of something like this in the future, but for now it's straight-forward to simply do it here (without going the de-tour of the extension library, waiting for a new Gardener release, and then call the introduced function).

rfranzke:
What's the status of this PR?

jia-jerry:
> What's the status of this PR?

Recently I'm handling some other urgent issues. Will continue next week. Thanks for your concern

rfranzke:
Hi @minchaow, I think this comment https://github.com/gardener/gardener-extension-provider-alicloud/pull/170#pullrequestreview-519455946 is not yet addressed in your PR.

Please only validate that the CIDR is not used if a new shoot is being created OR if the CIDR is trying to be changed in the update request.

rfranzke (approved):
/lgtm

Issue #173, opened 2020-11-06
LoadBalancer cannot be created


minchaow:
**How to categorize this issue?**

/area control-plane
/kind bug
/priority normal
/platform alicloud

**What happened**:
For `LoadBalancer` type service, the backend LoadBalancer cannot be created successfully.
**What you expected to happen**:
The backend LoadBalancer can be created successfully.
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
Error messages:

> Warning  CreatingLoadBalancerFailed  3m37s                    service-controller  Error creating load balancer (will retry): failed to ensure load balancer for service shoot--hc-dev--i302530-haas/kube-apiserver: Aliyun API Error: RequestId: FB167837-646D-4D84-9C0A-B01FC9A9CEE7 Status Code: 400 Code: ShareSlbHaltSales Message: The share instance has been discontinued.
>   Warning  CreatingLoadBalancerFailed  3m31s                    service-controller  Error creating load balancer (will retry): failed to ensure load balancer for service shoot--hc-dev--i302530-haas/kube-apiserver: Aliyun API Error: RequestId: D7FEB440-ECCD-4C97-A983-CC782DC0CC90 Status Code: 400 Code: ShareSlbHaltSales Message: The share instance has been discontinued.
>   Warning  CreatingLoadBalancerFailed  3m21s                    service-controller  Error creating load balancer (will retry): failed to ensure load balancer for service shoot--hc-dev--i302530-haas/kube-apiserver: Aliyun API Error: RequestId: 25AD93F5-DC89-41A9-91C3-49D16C4C66C4 Status Code: 400 Code: ShareSlbHaltSales Message: The share instance has been discontinued.
>   Normal   EnsuringLoadBalancer        3m1s (x4 over 3m37s)     service-controller  Ensuring load balancer
>   Warning  CreatingLoadBalancerFailed  3m                       service-controller  Error creating load balancer (will retry): failed to ensure load balancer for service shoot--hc-dev--i302530-haas/kube-apiserver: Aliyun API Error: RequestId: 6E19C658-3957-4552-91A5-56CD1C290081 Status Code: 400 Code: ShareSlbHaltSales Message: The share instance has been discontinued.
[...]

PR #174, opened 2020-11-06
mutate lb type service annotation

minchaow:

/area control-plane
/kind bug
/priority normal
/platform alicloud

This PR is to mutate `LoadBalancer` type service to contains `service.beta.kubernetes.io/alibaba-cloud-loadbalancer-spec: slb.s1.small` annotation, because alicloud ccm need this annotation when creating smal specification LoadBalancer.
Fixes #
#173
None

**Release note**:

```improvement operator

```


jia-jerry (dismissed):

pkg/webhook/utils/service.go
```diff
@@ -18,12 +18,16 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// MutateLBService mutates ServiceExternalTrafficPolicyTypeLocal of LoadBalancer type service
-func MutateLBService(new, old *corev1.Service) error {
+const (
+	AlicloudLoadBalancerSpecAnnotationKey = "service.beta.kubernetes.io/alibaba-cloud-loadbalancer-spec"
+	AlicloudLoadBalancerSmallSpec         = "slb.s1.small"
```
rfranzke, line 23: Out of curiosity/without understanding much about the contents of this PR, but why is this hard-coded here/does it make sense to make it configurable somewhere?

pkg/webhook/utils/service.go
```diff
@@ -32,6 +36,18 @@ func MutateLBService(new, old *corev1.Service) error {
...
+	if new.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		return
+	}
+
+	if new.Annotations == nil {
+		new.Annotations = make(map[string]string)
+	}
+	if _, exist := new.Annotations[AlicloudLoadBalancerSpecAnnotationKey]; !exist {
+		new.Annotations[AlicloudLoadBalancerSpecAnnotationKey] = AlicloudLoadBalancerSmallSpec
+	}
```
rfranzke, line 52: You could use `metav1.SetMetaDataAnnotation` like in https://github.com/gardener/gardener/blob/54311dcd102a42d0e851bc2a6d69797f4869ba1a/extensions/pkg/webhook/controlplane/checksums.go#L38

pkg/webhook/utils/service.go
minchaow: > Out of curiosity/without understanding much about the contents of this PR

It because now Alicloud needs the annotation `service.beta.kubernetes.io/alibaba-cloud-loadbalancer-spec` if we want to create `small` specification of loadbalancers, and this type loadbalancer is current default setting.

> but why is this hard-coded here/does it make sense to make it configurable somewhere?

Yes , I've already done some changes on the code, make it configuration from `configmap`.

pkg/webhook/utils/service.go
minchaow: done

charts/gardener-extension-provider-alicloud/values.yaml
```diff
@@ -74,6 +74,8 @@ config:
 #  - image-id1
 #  - image-id2
 #  ...
+  service:
+    backendLoadBalancerType: slb.s1.small
```
rfranzke, line 78: Shouldn't this be rather configured in the `CloudProfile` instead of in the extension config? 👀

minchaow: It's a good idea, but I have no idea how to leverage configure in `CloudProfile` in web hooks of the provider? Could you please provide one example for reference if possible? :)

rfranzke: Hm I understand, I don't see a straight-forward way either. OK, then ignore the question for now. thanks!

charts/gardener-extension-provider-alicloud/templates/configmap.yaml
```diff
@@ -25,7 +25,9 @@ data:
 {{- if .Values.config.kubeAPIServer }}
     kubeAPIServer:
       mutateExternalTrafficPolicy: {{ .Values.config.kubeAPIServer.mutateExternalTrafficPolicy }}
-{{- end}}
+{{- end }}
+    service:
+      backendLoadBalancerType: {{ .Values.config.service.backendLoadBalancerType }}
```
jia-jerry, line 30: Please rename backendLoadBalancerType to `backendLoadBalancerSpec` https://help.aliyun.com/document_detail/27577.html?spm=a2c4g.11174283.6.696.1c021192tmG2eY

charts/gardener-extension-provider-alicloud/values.yaml
jia-jerry, line 78: see above

hack/api-reference/config.md
```diff
@@ -275,4 +288,33 @@ bool
...
+<thead>
+<tr>
+<th>Field</th>
+<th>Description</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td>
+<code>backendLoadBalancerType</code></br>
```
jia-jerry, line 310: see above

pkg/apis/config/types.go
```diff
@@ -51,6 +53,11 @@ type KubeAPIServer struct {
 	MutateExternalTrafficPolicy bool
 }
 
+// LoadBalancerService specifies Service configuration
+type Service struct {
+	BackendLoadBalancerType string
```
jia-jerry, line 58: see above

jia-jerry (changes requested):

jia-jerry (approved):

PR #203, opened 2020-12-14
don't validate nil pod and service cidr

minchaow:

/area open-source
/kind bug
/priority normal
/platform alicloud

This PR is to remove validation of nil pod and service CIRD, otherwise it will block test machinery test cases, because Shoot manifests will be validated before the defaulter from `gardener-core-admission`.
Fixes #
None

**Release note**:

```other operator

```

### DIFF
--- a/pkg/apis/alicloud/validation/shoot.go
+++ b/pkg/apis/alicloud/validation/shoot.go
@@ -54,6 +54,10 @@ func ValidateNetworkingUpdate(newNetworking, oldNetworking core.Networking, fldP
 func ValidateNetworking(networking core.Networking, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	if networking.Nodes == nil {
+		allErrs = append(allErrs, field.Required(fldPath.Child("nodes"), "a CIDR must be provided for Alicloud shoots"))
+	}
+
 	allErrs = append(allErrs, validateNetworkCIDR(networking.Nodes, fldPath.Child("nodes"))...)
 	allErrs = append(allErrs, validateNetworkCIDR(networking.Pods, fldPath.Child("pods"))...)
 	allErrs = append(allErrs, validateNetworkCIDR(networking.Services, fldPath.Child("services"))...)
@@ -157,9 +161,7 @@ func validateVolumeFunc(volumeType *string, size string, fldPath *field.Path) fi
 func validateNetworkCIDR(cidr *string, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	if cidr == nil {
-		allErrs = append(allErrs, field.Required(fldPath, "a CIDR must be provided for Alicloud shoots"))
-	} else if cidrvalidation.NetworksIntersect(*cidr, ReservedCIDR) {
+	if cidr != nil && cidrvalidation.NetworksIntersect(*cidr, ReservedCIDR) {
 		allErrs = append(allErrs, field.Invalid(fldPath, fldPath, fmt.Sprintf("must not overlap with %s, it is reserved by Alicloud", ReservedCIDR)))
 	}
 
--- a/pkg/apis/alicloud/validation/shoot_test.go
+++ b/pkg/apis/alicloud/validation/shoot_test.go
@@ -77,14 +77,6 @@ var _ = Describe("Shoot validation", func() {
 					"Type":  Equal(field.ErrorTypeRequired),
 					"Field": Equal("spec.networking.nodes"),
 				})),
-				PointTo(MatchFields(IgnoreExtras, Fields{
-					"Type":  Equal(field.ErrorTypeRequired),
-					"Field": Equal("spec.networking.pods"),
-				})),
-				PointTo(MatchFields(IgnoreExtras, Fields{
-					"Type":  Equal(field.ErrorTypeRequired),
-					"Field": Equal("spec.networking.services"),
-				})),
 			))
 		})
 


jia-jerry (approved):

timebertt:
I'm a bit confused. There is probably a good reason for having this validation. 
Wouldn't it make more sense, to adapt the Shoot manifests used in TM tests to include the required values (cc @gardener/test-infra-maintainers)

minchaow:
> I'm a bit confused. There is probably a good reason for having this validation.
> Wouldn't it make more sense, to adapt the Shoot manifests used in TM tests to include the required values? (cc @gardener/test-infra-maintainers)

In my opinion, I think because we already have defaulter, so it doesn't matter to validate if these fields are nil or not.
If users miss these fields configuration, defaulter will add the default values; and if users configure these fields, validator will check if these values are intersected with the reserved CIDR.

Of course, adding required values in the Shoot manifests used in TM tests is also a good solution. :)

timebertt:
Hmm, I think you shouldn't rely on the defaulting logic in the Shoot API. If there is some special logic in the alicloud provider, that needs the CIDRs, then you should have the validation for it (so basically keep it).

jia-jerry:
> Hmm, I think you shouldn't rely on the defaulting logic in the Shoot API. If there is some special logic in the alicloud provider, that needs the CIDRs, then you should have the validation for it (so basically keep it).

The default logic is not from shoot API. If the network CIDRs are not set, default values from seed will be set. https://github.com/gardener/gardener/blob/master/plugin/pkg/shoot/validator/admission.go#L336. It is not integration testing shoot issue. Actually, shoot from dashboard has no CIDRs set for Service/Pods. The root cuase of this issue is that Gardener uses aggregated API server with admission plugins. Validation webhook takes effect *before* Gardener admission plugins. We should *NOT* check nil in the valdation webhook. At least for now.

minchaow:
> Hmm, I think you shouldn't rely on the defaulting logic in the Shoot API. If there is some special logic in the alicloud provider, that needs the CIDRs, then you should have the validation for it (so basically keep it).

@timebertt: hmm, it makes sense. OK, I will keep this part logic in the alicloud provider, but could you please help add required values in shoot manifest from test-machinery? We need run tests to validate our latest release. All tests are blocked now :(

minchaow:
I think this part of logic is necessary or not depends that `spec.networking.nodes`, `spec.networking.pods` and `spec.networking.services` are required values for shoot manifest or not?

jia-jerry:
> I think this part of logic is necessary or not depends that `spec.networking.nodes`, `spec.networking.pods` and `spec.networking.services` are required values for shoot manifest or not?

If you create a shoot from the dashboard, no cidrs are set for the 2 fields.

rfranzke:
Hm, actually, we don't have this kind of validation in AWS/Azure/GCP/etc. as well. I guess for the same reason: Option to let the pods/service CIDRs being defaulted. Probably we should change the validation to enforce the CIDRs when the `seedName` is not `nil`. WDYT?

timebertt:
> could you please help add required values in shoot manifest from test-machinery? We need run tests to validate our latest release. All tests are blocked now :(

@schrodit @dguendisch for help on this one :)

timebertt:
> Hm, actually, we don't have this kind of validation in AWS/Azure/GCP/etc. as well. I guess for the same reason: Option to let the pods/service CIDRs being defaulted. 

I'm not aware of the details, why such a validation in AliCloud was introduced in the first place. So, if it's not needed anymore just like in the other providers and it should just be defaulted, then we can drop it of course.

> Probably we should change the validation to enforce the CIDRs when the `seedName` is not `nil`. WDYT?

Hm, this can then lead to the scheduler's calls being blocked, right?

What I was rather asking myself is, why the validating webhooks are not called after the mutating in-tree admission plugins in g-api.
But it seems, we are just configuring it wrongly: https://github.com/gardener/gardener/blob/657ee148edb51f09d0571441abaaa56fcf0c5657/cmd/gardener-apiserver/app/gardener_apiserver.go#L187-L206
We should configure g-api to invoke the `ValidatingAdmissionWebhook` admission plugin after the in-tree `ShootValidator` plugin.

rfranzke:
> Hm, this can then lead to the scheduler's calls being blocked, right?

Yes, but that's the desired behaviour then. When a shoot is assigned to a seed then the Alicloud (and most other providers) require pods/services to be set.

> We should configure g-api to invoke the ValidatingAdmissionWebhook admission plugin after the in-tree ShootValidator plugin.

Yep, this makes also sense 👍

dguendisch:
> @schrodit @dguendisch for help on this one :)

Sorry, but I didn't get so far what is the desired state. We intentionally removed POD and svc CIDRs from the shoot templates about two months ago so that alicloud can default to their own CIDRs which are different from other cloudproviders (@jia-jerry  I guess you remember): https://github.wdf.sap.corp/kubernetes/landscape-setup/commit/b04c3c44f7f5b06cb816107f73ebf500d0379f63#diff-25eab7af54af12394eb933cde8bbe861

Of course that change can be reverted/customized to some different CIDRs, is it that what you @minchaow want? Or you want to first clarify more with the colleagues about the correct target picture?

jia-jerry:
> > @schrodit @dguendisch for help on this one :)
> 
> Sorry, but I didn't get so far what is the desired state. We intentionally removed POD and svc CIDRs from the shoot templates about two months ago so that alicloud can default to their own CIDRs which are different from other cloudproviders (@jia-jerry I guess you remember): https://github.wdf.sap.corp/kubernetes/landscape-setup/commit/b04c3c44f7f5b06cb816107f73ebf500d0379f63#diff-25eab7af54af12394eb933cde8bbe861
> 
> Of course that change can be reverted/customized to some different CIDRs, is it that what you @minchaow want? Or you want to first clarify more with the colleagues about the correct target picture?

jia-jerry:
> > @schrodit @dguendisch for help on this one :)
> 
> Sorry, but I didn't get so far what is the desired state. We intentionally removed POD and svc CIDRs from the shoot templates about two months ago so that alicloud can default to their own CIDRs which are different from other cloudproviders (@jia-jerry I guess you remember): https://github.wdf.sap.corp/kubernetes/landscape-setup/commit/b04c3c44f7f5b06cb816107f73ebf500d0379f63#diff-25eab7af54af12394eb933cde8bbe861
> 
> Of course that change can be reverted/customized to some different CIDRs, is it that what you @minchaow want? Or you want to first clarify more with the colleagues about the correct target picture?

Please hold on. I think leave it blank is reasonable, just like creating a shoot from dashboard.

jia-jerry:
> What I was rather asking myself is, why the validating webhooks are not called after the mutating in-tree admission plugins in g-api.
> But it seems, we are just configuring it wrongly: https://github.com/gardener/gardener/blob/657ee148edb51f09d0571441abaaa56fcf0c5657/cmd/gardener-apiserver/app/gardener_apiserver.go#L187-L206
> We should configure g-api to invoke the `ValidatingAdmissionWebhook` admission plugin after the in-tree `ShootValidator` plugin.

It will be great if we could achieve this. But how can we do that?

timebertt:
> It will be great if we could achieve this. But how can we do that?

Will work on it now :)

jia-jerry:
> > Hm, this can then lead to the scheduler's calls being blocked, right?
> 
> Yes, but that's the desired behaviour then. When a shoot is assigned to a seed then the Alicloud (and most other providers) require pods/services to be set.
> 
> > We should configure g-api to invoke the ValidatingAdmissionWebhook admission plugin after the in-tree ShootValidator plugin.
> 
> Yep, this makes also sense 👍

So shall we change the order of validation. Though from my understanding, 

> > It will be great if we could achieve this. But how can we do that?
> 
> Will work on it now :)

Thank you very much. I thought webhook is always called by kube-apiserver.

timebertt:
Ok, the order in g-api is now corrected.
If I got it correctly, we don't need to remove the CIDR validation here, right?

However, when implementing it, I noticed, that the networking defaulting was anyways happening before the validatingwebhook call, because the `Shoot` admission plugin is actually a mutating plugin (mutating plugins are invoked before all validating plugins by default already).
So there must be some other problem in the TM tests...

jia-jerry:
> Ok, the order in g-api is now corrected.
> If I got it correctly, we don't need to remove the CIDR validation here, right?
> 
> However, when implementing it, I noticed, that the networking defaulting was anyways happening before the validatingwebhook call, because the `Shoot` admission plugin is actually a mutating plugin (mutating plugins are invoked before all validating plugins by default already).
> So there must be some other problem in the TM tests...

After g-api is corrected, this PR is not needed. The default CIDR is from seed. As the seed for Alicloud is well configured, the shoot should be valid one. I think TM test shoots will also be OK. Thank you for your support.

PR #205, opened 2020-12-15
Revert "Forbid network cidr 100.64.0.0/10"

jia-jerry:
Reverts gardener/gardener-extension-provider-alicloud#170

By default, a shoot created from dashboard has no pod/service network settings. It will fail to pass validation check, though there is a defaulter logic in shoot admission in g/g. The issue is that the validation webhook takes place before admission plugins in g/g. https://github.com/gardener/gardener/pull/3298 fixed the order issue. We will dismiss this revert after fix in g/g is released.


minchaow (approved):
/lgtm

PR #206, opened 2020-12-16
Pull CSI driver only if it is not present

ialidzhikov:
/area cost
/platform alicloud

**Release note**:

```other user
The imagePullPolicy of the Alicloud CSI driver is now switched from `Always` to `IfNotPresent`. This should avoid unnecessary image pulls when the image is already present.
```


rfranzke (approved):
/lgtm

minchaow (approved):
/lgtm

PR #207, opened 2020-12-16
Update kubernetes-csi/external-snapshotter to v2.1.3

ialidzhikov:
/area storage
/platform alicloud

**Release note**:

```other user
The following images are updated to address [CVE-2020-8569](https://groups.google.com/g/kubernetes-dev/c/_2a2BVACg3s/m/taaiikXRAgAJ):
- quay.io/k8scsi/csi-snapshotter: v2.1.1 -> v2.1.3
- quay.io/k8scsi/snapshot-controller: v2.1.1 -> v2.1.3
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -54,12 +54,12 @@ images:
 - name: csi-snapshotter
   sourceRepository: https://github.com/kubernetes-csi/external-snapshotter
   repository: quay.io/k8scsi/csi-snapshotter
-  tag: v2.1.1
+  tag: v2.1.3
   targetVersion: ">= 1.17"
 - name: csi-snapshot-controller
   sourceRepository: https://github.com/kubernetes-csi/external-snapshotter
   repository: quay.io/k8scsi/snapshot-controller
-  tag: v2.1.1
+  tag: v2.1.3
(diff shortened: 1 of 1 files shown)


rfranzke (approved):
/lgtm

minchaow (approved):
/lgtm
but I suppose we still need to validate this update of CSI plugins. I am wondering if the tests from test machinery will cover this part?

ialidzhikov:
> but I suppose we still need to validate this update of CSI plugins. I am wondering if the tests from test machinery will cover this part?

I think `CSI plugin` rather refers to the cloud provider specific CSI driver. In this PR we don't update the CSI driver. We only update the the `csi-snapshotter` sidecar and the `snapshot-controller`. However as always, you are more than welcome to play with it locally.
Wrt to the test-machinery, I don't think that we currently run VolumeSnapshot specific tests as part of the conformance tests. To my knowledge there is this PR https://github.com/gardener/test-infra/pull/311 in test-infra to enable more CSI tests but it is not hold since quite some time and I am sure whether it will cover VolumeSnapshot lifecycle operations. 

/cc @dguendisch

PR #212, opened 2020-12-19
Check the csi-snapshotter-controller only for >= 1.17

ialidzhikov:
/kind bug

**Release note**:

```bugfix operator
An issue causing provider-alicloud to wrongly report failed health check because of missing csi-snapshot-controller Deployment for Kubernetes < 1.17 is now fixed. csi-snapshot-controller Deployment is enabled for Kubernetes >= 1.17.
```


timebertt (approved):
/lgtm

PR #221, opened 2021-01-08
modify logic of networking validation

minchaow:

/area networking
/kind enhancement
/priority normal
/platform alicloud

This PR did some changes on the logic of networking CIDRs validations.

Pods and Services' CIDRs would not be defaulted if `spec.seedName` is nil, because of the code line in the file https://github.com/gardener/gardener/blob/ec5794c5fddf09388f4e976f52327fd3d3d152b1/plugin/pkg/shoot/validator/admission.go#L336.
```
if seed != nil {
  ...
}
```
Normally `spec.seedName` in shoot is nil, and is set by `gardener-scheduler` once one seed is picked, and it is invoked after the alicloud validator. So the validation logic for the shoot creation is as follow:
1. Allow the request when seedName and pods and services' CIDRs are unset.
2. Only deny the request when the CIDRs are set to reserved CIDR.

Besides, this PR also did some changes on the validation of immutable fields logic during shoot updates. 

Shoots' update requests will be sent to the gardener apiserver after the seed name is set. If alicloud validator blocks all update requests for networking, CIDRs cannot be defaulted by the gardener admission. So the validation logic for the shoot update is as follow:
1. Allow the update from invalidated CIDR to validated CIDR ("null" => "127.0.0.1/16").
1. Deny the update from validated CIDR to validated CIDR ("127.0.0.1/16" => "120.0.0.1/24").

Fixes #
None

**Release note**:

```other operator

```

### DIFF
--- a/pkg/apis/alicloud/validation/shoot.go
+++ b/pkg/apis/alicloud/validation/shoot.go
@@ -16,6 +16,7 @@ package validation
 
 import (
 	"fmt"
+	"net"
 	"regexp"
 
 	apisalicloud "github.com/gardener/gardener-extension-provider-alicloud/pkg/apis/alicloud"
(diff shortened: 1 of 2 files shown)


timebertt (approved):
Great, @minchaow! This is exactly the logic I was thinking of :)
/lgtm

PR #233, opened 2021-02-03
refactor alicloud clients

minchaow:

/area dev-productivity
/kind enhancement
/priority 3
/platform alicloud

This PR is to do some refactoring of the structure of Alicloud clients.
1. Embed alicloud clients into implemented structs like `ecsClient`, `vpcClient`, `slbClient`, `ramClient`, `stsClient` and `ossClient`.
2. We still don't expose these structs and only expose interface like `ECS`, `VPC`, `RAM` and so on.  Because alicloud doesn't leverage interface to implement their clients' methods, so we have to add the same methods in our interface but don't need to implement(Alicloud clients are embedded in our struct, and it already implemented the interface), like as follow:
```
type VPC interface {
        ... ...
        // Only need add the function into the interface, no need to implement it.
	DescribeVSwitches(request *vpc.DescribeVSwitchesRequest) (response *vpc.DescribeVSwitchesResponse, err error)
	... ...
}

clientFactory.NewVPCClient().DescribeVpcs(request)
```
3. All declared clients are created from the `clientFactory` now.

Fixes #
None

**Release note**:

```other operator
None
```


pkg/alicloud/client/types.go
```diff
@@ -34,47 +38,89 @@ type ClientFactory interface {
...
-	DeleteObjectsWithPrefix(ctx context.Context, bucketName, prefix string) error
-	CreateBucketIfNotExists(ctx context.Context, bucketName string) error
-	DeleteBucketIfExists(ctx context.Context, bucketName string) error
+// ramClient implements the interface of RAM.
+type ramClient struct {
+	ram.Client
 }
 
-// RAM is an interface which must be implemented by alicloud ram clients.
+// RAM is an interface which provides RAM related methods.
```
jia-jerry, line 110:
```suggestion
// RAM is an interface which declares RAM related methods.
```

pkg/controller/infrastructure/client.go
```diff
@@ -15,89 +15,68 @@
...
 	}, nil
 }
 
 // FetchEIPInternetChargeType fetches the internet charge type for the VPC's EIP.
-func FetchEIPInternetChargeType(vpcClient alicloudclient.VPC, vpcID string) (string, error) {
-	describeNATGatewaysReq := vpc.CreateDescribeNatGatewaysRequest()
-	describeNATGatewaysReq.VpcId = vpcID
-
-	describeNatGatewaysRes, err := vpcClient.DescribeNatGateways(describeNATGatewaysReq)
+func FetchEIPInternetChargeType(ctx context.Context, vpcClient alicloudclient.VPC, vpcID string) (string, error) {
```
jia-jerry, line 59: Make it private and as an instance function of vpcClient if it is not used in other places.  This method is called `GetVPCInfo`, where NATGateway has been retrieved. Please just try NATGateway as an input parameter.

pkg/controller/infrastructure/client.go
```diff
@@ -15,89 +15,68 @@
...
+	natGateway, err := vpcClient.GetNatGatewayWithVPCID(ctx, vpcID)
```
jia-jerry, line 60: The interface name is tricky. I suppose it will return a `list` of natGatway. 
```suggestion
	natGateways, err := vpcClient.GetNatGatewayWithVPCID(ctx, vpcID)
```

jia-jerry (changes requested):

pkg/alicloud/client/types.go
```diff
@@ -34,47 +38,89 @@ type ClientFactory interface {
 	NewSLBClient(region, accessKeyID, accessKeySecret string) (SLB, error)
 	NewVPCClient(region, accessKeyID, accessKeySecret string) (VPC, error)
 	NewRAMClient(region, accessKeyID, accessKeySecret string) (RAM, error)
-	NewStorageClientFromSecretRef(ctx context.Context, client client.Client, secretRef *corev1.SecretReference, region string) (Storage, error)
+	NewOSSClient(endpoint, accessKeyID, accessKeySecret string) (OSS, error)
+	NewOSSClientFromSecretRef(ctx context.Context, client client.Client, secretRef *corev1.SecretReference, region string) (OSS, error)
```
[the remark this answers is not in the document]
minchaow: it was call inside the function `NewOSSClientFromSecretRef`.

pkg/alicloud/client/types.go
```diff
@@ -34,47 +38,89 @@ type ClientFactory interface {
...
 }
 
-// ECS is an interface which must be implemented by alicloud ecs clients.
+// ECS is an interface which provides ECS related methods.
 type ECS interface {
 	CheckIfImageExists(ctx context.Context, imageID string) (bool, error)
 	ShareImageToAccount(ctx context.Context, regionID, imageID, accountID string) error
+	DescribeSecurityGroups(ctx context.Context, request *ecs.DescribeSecurityGroupsRequest) (response *ecs.DescribeSecurityGroupsResponse, err error)
+	DescribeSecurityGroupAttribute(ctx context.Context, request *ecs.DescribeSecurityGroupAttributeRequest) (response *ecs.DescribeSecurityGroupAttributeResponse, err error)
+	DescribeKeyPairs(ctx context.Context, request *ecs.DescribeKeyPairsRequest) (response *ecs.DescribeKeyPairsResponse, err error)
```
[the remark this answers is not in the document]
minchaow: these functions are referenced in the infra test integration, see https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/test/integration/infrastructure/infrastructure_test.go

pkg/alicloud/client/types.go
```diff
@@ -34,47 +38,99 @@ type ClientFactory interface {
...
 }
 
-// SLB is an interface which must be implemented by alicloud slb clients.
+// stsClient implements the STS interface.
+type stsClient struct {
```
jia-jerry, line 60: We do  not need a wrapper struct. We just need to return an instance of `sts.Client` for those NewSTSClient. The same for other wrappers.

pkg/alicloud/client/types.go
```diff
@@ -34,47 +38,99 @@ type ClientFactory interface {
...
+	sts.Client
```
jia-jerry, line 61: We do  not need a wrapper struct. We just need to return an instance of `sts.Client` for those NewSTSClient. The same for other wrappers.

jia-jerry (changes requested):

pkg/alicloud/client/types.go
minchaow: stsClient as well as other wrappers, not only would be used to call interface provided by Alicloud, but also implements the interface which defined by us. So we need this kind of wrapper.

jia-jerry (approved):

PR #237, opened 2021-02-18
remove shared vpc for integration tests

minchaow:

/area testing
/kind enhancement
/priority normal
/platform alicloud

This PR is to remove the usage of shared VPC for infra integration tests. Because enhanced type NAT on Alicloud is supported now, and its cost is much cheaper than before. So it is not necessary to leverage shared VPC for testing.
Fixes #
None
None
**Release note**:

```other developer
Shared VPC and NATs are removed for integration tests.
```

### DIFF
--- a/.ci/testruns/default/templates/testrun.yaml
+++ b/.ci/testruns/default/templates/testrun.yaml
@@ -37,9 +37,3 @@ spec:
       - name: REGION
         type: env
         value: eu-central-1
-      - name: VPC_ID
-        type: env
-        valueFrom:
-          secretKeyRef:
-            name: alicloud-configuration
-            key: vpcID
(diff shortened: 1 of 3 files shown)


minchaow:
Hi @schrodit I did some modification for the integration testing and removed `vpc-id` from the parameters, are there any other changes required for test-machinery side?

schrodit:
@dguendisch  I think we already adjusted the testruns accordingly

dguendisch:
Yes, from testmachinery side the shared VPC is already removed, so nothing to be done from testmachinery side.

minchaow:
@schrodit @dguendisch thanks for your inputs.:)

jia-jerry (dismissed):

jia-jerry (approved):

PR #246, opened 2021-03-02
Allow MCM to list and watch PDBs

ialidzhikov:
/kind bug

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/machine-controller-manager/shoot/templates/clusterrole-machine-controller-manager.yaml
+++ b/charts/internal/machine-controller-manager/shoot/templates/clusterrole-machine-controller-manager.yaml
@@ -60,3 +60,10 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - policy
+  verbs:
+  - list
+  - watch
+  resources:
+  - poddisruptionbudgets


minchaow (commented):
Thanks for your fix so much.
I am wondering about the permission in the following two files https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/charts/internal/machine-controller-manager/seed/templates/clusterrole.yaml and  https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/charts/internal/machine-controller-manager/shoot/templates/role-machine-controller-manager.yaml, do they also need permission of PDBs?

vpnachev (approved):
/lgtm

ialidzhikov:
> Thanks for your fix so much.
> I am wondering about the permission in the following two files https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/charts/internal/machine-controller-manager/seed/templates/clusterrole.yaml and https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/charts/internal/machine-controller-manager/shoot/templates/role-machine-controller-manager.yaml, do they also need permission of PDBs?

Nope

PR #278, opened 2021-05-04
fix nil pointer issue

minchaow:

/area robustness
/kind bug
/platform alicloud

This PR is to fix the panic issue of `nil` pointer.
Fixes #
None
None
**Release note**:

```other operator

```


jia-jerry (approved):

minchaow:
@rfranzke @timebertt I checked the CI pipelines, the image `build_oci_image_gardener-extension-validator-alicloud` seems would not be built any more, so do we need do some changes for github checks?

rfranzke:
@minchaow done

PR #64, opened 2020-04-06
Remove zone of controlplane config

jia-jerry:
**What this PR does / why we need it**:
1. Remove zone field in shoot controlPlaneConfig section
2. Change Cloud controller configmap used by CCM to secret
**Which issue(s) this PR fixes**:
Fixes #63 

**Special notes for your reviewer**:

**Release note**:

```improvement operator
Zone filed of controlPlaneConfig is removed from Shoot.
```
```improvement user
Now user does not need to maintain Zone filed of controlPlaneConfig in Shoot resource. And also, controlPlaneConfig is not needed.
```

PR #279, opened 2021-05-04
add new field in controlplaneconfig

minchaow:

/area control-plane
/kind bug
/platform alicloud

This PR is to fix the issue that failing to decode `ControlPlaneConfig` after strict decoder is enabled.
```
controlPlaneConfig:
  apiVersion: alicloud.provider.extensions.gardener.cloud/v1alpha1
  kind: ControlPlaneConfig
  zone: cn-shanghai-e
```

Currently `zone` under `controlPlaneConfig` is defaulted by dashboard template of alicloud.

Error message is as follow:

> strict decoder error for {"apiVersion":"alicloud.provider.extensions.gardener.cloud/v1alpha1","kind":"ControlPlaneConfig","zone":"cn-shanghai-e"}: v1alpha1.ControlPlaneConfig.ReadObject: found unknown field: zone, error found in #10 byte of ...|ig","zone":"cn-shang|..., bigger context ...|cloud/v1alpha1","kind":"ControlPlaneConfig","zone":"cn-shanghai-e"}|... stacktrace:github.com/go-logr/zapr.(*zapLogger).Error

Fixes #
None
None
**Release note**:

```other operator
None
```

### DIFF
--- a/hack/api-reference/api.md
+++ b/hack/api-reference/api.md
@@ -97,6 +97,18 @@ string
 </tr>
 <tr>
 <td>
+<code>zone</code></br>
+<em>
+string
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>Zone is the zone ID of the control plane.</p>
+</td>
+</tr>
+<tr>
+<td>
 <code>cloudControllerManager</code></br>
 <em>
 <a href="#alicloud.provider.extensions.gardener.cloud/v1alpha1.CloudControllerManagerConfig">
--- a/pkg/apis/alicloud/types_controlplane.go
+++ b/pkg/apis/alicloud/types_controlplane.go
@@ -24,6 +24,9 @@ import (
 type ControlPlaneConfig struct {
 	metav1.TypeMeta
 
+	// Zone is the zone ID of the control plane.
+	Zone *string
+
 	// CloudControllerManager contains configuration settings for the cloud-controller-manager.
 	CloudControllerManager *CloudControllerManagerConfig
 }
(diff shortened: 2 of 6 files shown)


ialidzhikov:
> Currently `zone` under `controlPlaneConfig` is defaulted by dashboard template of alicloud.

Why you re-introduce this field (that was removed with https://github.com/gardener/gardener-extension-provider-alicloud/pull/64) instead of fixing the wrong dashboard template?

ialidzhikov:
And the ControlPlaneConfig in the dashboard seems fine to me - https://github.com/gardener/dashboard/blob/833be8afbf0235426f2eff45ae9a786e1bd8bcd9/frontend/src/utils/createShoot.js#L143-L146.

minchaow:
> Why you re-introduce this field (that was removed with #64) instead of fixing the wrong dashboard template?

Current situation is a bit complex. The dashboard template was released a bit early, so some shoots contain this field but some don't. In order to make it compatible with all shoots, so I re-introduce this field, but set it as optional.

ialidzhikov:
I don't get, this field is removed since 1 year and I don't see any recent changes to the dashboard template and the dashboard template seems fine to me.

ialidzhikov:
> I don't get, this field is removed since 1 year and I don't see any recent changes to the dashboard template and the dashboard template seems fine to me.

Ups, I was wrong here, I now created a PR that stops setting the zone field - https://github.com/gardener/dashboard/pull/1009.

minchaow:
this field introduced in dashboard template in the code https://github.com/gardener/dashboard/blob/833be8afbf0235426f2eff45ae9a786e1bd8bcd9/frontend/src/utils/createShoot.js#L297-L310, will be fixed by the PR https://github.com/gardener/dashboard/pull/1009

minchaow:
close the PR due to https://github.com/gardener/gardener-extension-provider-alicloud/pull/280

PR #280, opened 2021-05-04
disable strict mode of decoder

minchaow:

/area control-plane
/kind bug
/platform alicloud

This PR is to hack decoder of `ValuesProvider`, to leverage one custom-decoder (strict mode disabled).
The `Zone` field in not required in "ControlPlaneConfig" any more, and it has already been removed in the struct, but still exists in shoots' yaml. So it should also be removed in the shoots' yaml in the future.

Here we leverage one custom decoder (disabled the strict mode) to make migration more smooth.

Related change on dashboard, see the PR https://github.com/gardener/dashboard/pull/1009.

Fixes #
None
None
**Release note**:

```bugfix user
The field "zone" under the filed "ControlPlaneConfig" is not required any more, shoots' users should remove is manually.
```


rfranzke (approved):
/lgtm

PR #282, opened 2021-05-07
Revert "[ci:component:github.com/gardener/terraformer:v2.5.0->v2.6.0]"

minchaow:
Reverts gardener/gardener-extension-provider-alicloud#277

With TF provider of version >=1.119 , one new field named `enable_ipv6` would be added into the VPC resource, which will trigger a force replacement. It only impacts the VPC which has already been existing for a long time.

It its blocking the release of extension-provider-alicloud, so revert the change and will restore it when the issue is fixed.

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -2,7 +2,7 @@ images:
 - name: terraformer
   sourceRepository: github.com/gardener/terraformer
   repository: eu.gcr.io/gardener-project/gardener/terraformer-alicloud
-  tag: "v2.6.0"
+  tag: "v2.5.0"
 - name: machine-controller-manager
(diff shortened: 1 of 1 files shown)


timebertt (approved):
/lgtm

Just for my understanding: do you also plan to revert https://github.com/gardener/terraformer/pull/91 or what fix do you have in mind?

minchaow:
> /lgtm
> 
> Just for my understanding: do you also plan to revert [gardener/terraformer#91](https://github.com/gardener/terraformer/pull/91) or what fix do you have in mind?

I don't want to revert it to the old version again. When the issue is fix from Alicloud, I will update the it to the newer version.

rfranzke (approved):

PR #286, opened 2021-05-11
Do not use strict decoding for ControlPlaneConfig in admission-alicloud

ialidzhikov:
/kind bug
/platform alicloud

Similar to https://github.com/gardener/gardener-extension-provider-alicloud/pull/280.

admission-alicloud should be able to tolerate the `zone` field to make possible updating existing Shoots that still have the `zone` field. 

**Release note**:

```other operator
NONE
```


minchaow (approved):
/lgtm

PR #310, opened 2021-06-15
Use separate ManagedResource for ControlPlane CRDs

ialidzhikov:
/area storage
/kind enhancement
/platform alicloud

Similar to https://github.com/gardener/gardener-extension-provider-aws/pull/337

Part of gardener/gardener#2227

**Release note**:

```bugfix operator
provider-alicloud is now using a separate ManagedResource for ControlPlane CRDs (`volumesnapshot` related CRDs) that are installed in the Shoot cluster to separate the deletion of CRDs from the deletion of the RBAC for controller leader election.
```

```breaking operator
This version of provider-alicloud requires at least Gardener v1.21.0. Before upgrading to this version of provider-alicloud, make sure that you upgraded to at least Gardener v1.21.0.
```

### DIFF
--- a/charts/internal/shoot-crds/Chart.yaml
+++ b/charts/internal/shoot-crds/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for control plane CRDs resources that are installed in the Shoot cluster
+name: shoot-crds
+version: 0.1.0
--- a/charts/internal/shoot-crds/charts/volumesnapshots/Chart.yaml
+++ b/charts/internal/shoot-crds/charts/volumesnapshots/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for volumesnapshot CRDs.
+name: volumesnapshots
+version: 0.1.0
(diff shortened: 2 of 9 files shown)


charts/internal/shoot-system-components/charts/csi-alicloud/templates/crd-volumesnapshots.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   annotations:
     api-approved.kubernetes.io: "https://github.com/kubernetes-csi/external-snapshotter/pull/260"
     resources.gardener.cloud/delete-on-invalid-update: "true"
```
rfranzke, line 9: Maybe this should be removed? 👀

rfranzke (approved):
/lgtm

charts/internal/shoot-system-components/charts/csi-alicloud/templates/crd-volumesnapshots.yaml
ialidzhikov: This annotation was introduced to make the update from K8s v1.16 to v1.17 (from external-snapshotter@v1 to external-snapshotter@v2) possible. If we remove it now, I guess this will regress such update. I now created https://github.com/gardener/gardener-extension-provider-alicloud/issues/314 with more detailed description. So I guess the safer option is to drop support for v1.15.x and v1.16.x and then remove this annotation from the volumesnapshot CRDs. WDYT?

rfranzke: Sorry for the late reply. Sure, I'm fine. Thanks!

PR #315, opened 2021-06-22
Not to check valid values for immutable fields when updating of a shoot

jia-jerry:

/area control-plane
/kind enhancement
/platform alicloud

We don't validate values for pod/service/node network. This is for compatibility consideration to avoid user can't delete a shoot which has invalid network settings.
Fixes #

**Release note**:

```other operator
Users can delete a shoot with invalid network settings.
```

### DIFF
--- a/pkg/admission/validator/shoot.go
+++ b/pkg/admission/validator/shoot.go
@@ -92,10 +92,6 @@ func (s *shoot) Validate(ctx context.Context, new, old client.Object) error {
 }
 
 func (s *shoot) validateShoot(_ context.Context, shoot *core.Shoot, infraConfig *alicloud.InfrastructureConfig, cpConfig *alicloud.ControlPlaneConfig) error {
-	// Network validation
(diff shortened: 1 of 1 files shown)


rfranzke:
Now you make assumptions about the internal validation code of g/g which might change in the future and break your assumptions. Proposal: Check yourself whether the `networking` section changed and only perform validation if there is a difference.

jia-jerry:
> Now you make assumptions about the internal validation code of g/g which might change in the future and break your assumptions. Proposal: Check yourself whether the `networking` section changed and only perform validation if there is a difference.

Not really. Immutable constraints are applied in this extension: https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/pkg/apis/alicloud/validation/shoot.go#L47

minchaow (approved):
/lgtm

minchaow:
If I understand correctly, we don't need validate the CIDR for updating shoots, because its value is immutable?

rfranzke (approved):

PR #323, opened 2021-07-06
Improve docs/usage-as-end-user.md

ialidzhikov:
This PR:
- Moves the Secret format/example from the `Permissions` section to the `Alicloud Provider Credentials` section (similar to what we have in provider-aws)
- Makes the Alicloud RAM policy hidden by default and expandable on demand (similar to what we have in provider-aws)

**Release note**:

```other operator
NONE
```


rfranzke (approved):
/lgtm

PR #325, opened 2021-07-13
Add DNSRecord controller

stoyanr:

/area control-plane
/kind api-change
/kind enhancement
/platform alicloud

* Adds a Alicloud-specific implementation of a DNSRecord controller, see https://github.com/gardener/gardener/pull/4231.
* Vendors g/g v1.27.0-dev and k8s v0.21.2 (in a separate commit to make reviews easier)

Fixes #

Note the following important details:
* The provider type is `alicloud-dns` (like in `DNSProvider`), not `alicloud` (like in other controllers) for reasons of compatibility and ease of migration.

See also https://github.com/gardener/gardener-extension-provider-aws/pull/362 and https://github.com/gardener/gardener-extension-provider-aws/pull/370.

**Release note**:

```other operator
NONE
```


stoyanr:
Unlike `aws` and `azure`, creating / updating and deleting DNS records involves first getting existing DNS records with this name and type from the zone (or "domain" in Alicloud terminology). Since records in Alicloud have IDs, this is the only reliable way to implement it so that it behaves correctly with any actual state, and consistently with other provider implementations. 

This does however result in more requests to the DNS provider API in some cases
* Creating a new record does `get + create`, instead of just `createOrUpdate`
* Deleting a record does `get + delete`, instead of just `delete`
* Updating an existing record is usually just `get` (when there are no changes, which is 99.99% of the cases), instead of `createOrUpdate`, so there is no change of the number of requests in this case

I considered getting the record IDs and saving them in the resource `status` so that I could use them for update / deletion, but then realized that:
* this doesn't avoid the `get` in the creation case (since the records might exist there already, and we can't know the IDs)
* it would not handle properly any external changes to the zone (e.g. a record was deleted and then created manually)
 
So eventually, I favored the most consistent and "foolproof" implementation that is also consistent with the behavior of the other provider extensions and also the current `dns-external` controller, even though it could result in more requests to the provider API in some case. Since the Alicloud DNS is not heavily used by gardener shoots, I expect that rate limits here are not much of an issue.

This also has been discussed and agreed with @timebertt  and @plkokanov.

pkg/alicloud/secret.go
```diff
@@ -35,22 +36,32 @@ const (
 	AccessKeyID = "accessKeyID"
 	// AccessKeySecret is the data field in a secret where the access key secret is stored at.
 	AccessKeySecret = "accessKeySecret"
+
+	// DNSAccessKeyID is the data field in a DNS secret where the access key id is stored at.
+	DNSAccessKeyID = "ACCESS_KEY_ID"
```
jia-jerry, line 41: I would suggest to use dnsAccessKeyID / dnsAccessKeySecret.

pkg/alicloud/secret.go
```diff
@@ -60,11 +71,23 @@ func ReadSecretCredentials(secret *corev1.Secret) (*Credentials, error) {
 }
 
 // ReadCredentialsFromSecretRef reads the credentials from the secret referred by given <secretRef>.
-func ReadCredentialsFromSecretRef(ctx context.Context, client client.Client, secretRef *corev1.SecretReference) (*Credentials, error) {
+func ReadCredentialsFromSecretRef(ctx context.Context, client client.Client, secretRef *corev1.SecretReference, allowDNSKeys bool) (*Credentials, error) {
```
jia-jerry, line 74: I would suggest introduce another function to read DNS credentials only. The name is like `ReadDNSCredentialsFromSecretRef`.The function tries  to read `dnsAccessKeyID / dnsAccessKeySecret` first. If they don't exist, tries `accessKeyID / accessKeySecret` as fallback.

pkg/alicloud/secret.go
stoyanr: Done

pkg/alicloud/secret.go
stoyanr: I introduced the function `ReadDNSCredentialsFromSecretRef` as you suggested, but the only difference from `ReadCredentialsFromSecretRef` is that is passes `true` instead of false as the last argument to `ReadSecretCredentials`, in this way allowing DNS secret keys in addition to the non-DNS ones. I didn't implement checking for the DNS key first though, since this would mean introducing additional complexity in `ReadSecretCredentials` (or duplicating most of the logic in another function), and the difference would be really minor - both variants of secret keys are already supported, and the order in which they are checked would only matter if we have both e.g. `accessKeyID` and `ACCESS_KEY_ID` in the same secret, which I don't consider to be likely or important enough to make the logic more complex here.

stoyanr: Also have in mind that this logic is very similar to what's implemented in all other provider extensions, and it's already merged (or going to be merged soon).

stoyanr:
@jia-jerry Thanks for your suggestions, I addressed them, could you please take another look?

pkg/alicloud/secret.go
jia-jerry: Thanks. You are right. I checked the code again. The secret used by DNS is different with that used by infra.

jia-jerry: Please just keep the same logic as other providers. And also, I think `ReadDNSCredentialsFromSecretRef` is not needed.

stoyanr: Since I've already introduced the new function, let's keep it, it's a rather cosmetic change and it's ok this way I think.

jia-jerry (approved):
/ lgtm

PR #333, opened 2021-07-30
Include encrypted field of volume when calculating worker hash

jia-jerry:

/area control-plane
/kind bug
/platform alicloud

When enduser sets workers[].volume.encrypted=true, we need to create a new MachineClass so that related nodes will be recreated.
Fixes #

**Release note**:

```bugfix user
When user set encrypted value for volume in workers, all affected nodes will be recreated. 
```


vpnachev (commented):
Can you add a unit test about this change?

jia-jerry:
> Can you add a unit test about this change?

done.

EmoinLanyu (approved):

PR #336, opened 2021-08-03
add deploy infra task when encrypted disk is enabled

### LINKED ISSUE #335: Change of system disk encrypted option doesn't trigger node update
**How to categorize this issue?**

/area control-plane
/kind enhancement
/priority 3
/platform alicloud

**What happened**:

When user changed a worker of shoot with encrypted system volume, the shoot worker will be reconciled with failure. 
The root cause is the `infra` is not reconciled, during which, the encrypted custom image is prepared in infra status. Currently, as a workaround, user have to `maintain` the shoot.
**What you expected to happen**:
If a worker's `encrypted system volume` is changed, the infra will be reconciled automatically.
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Gardener version (if relevant):
- Extension version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

jia-jerry:

/area control-plane
/kind enhancement
/platform alicloud

Fixes #335 

In AliCloud, to support encrypted system disk, we have to make an encrypted customer VM image based on an existing plain one. Then we use this encrypted image to deploy a machine.

In the infra reconciliation, we check if there already is an encrypted image. Otherwise, in the controller loop, it will create an encrypted image. If the encrypted image is ready, the image id will be put in  the status of the infrastructure resource.

In the worker reconciliation phase, the controller will find the suitable encrypted image id from the infra status. If it doesn't exist, the worker reconciliation will fail.

**Release note**:

```feature operator
When user changes system volume to encrypted in a shoot, infrastructure will be reconciled automatically so that workers can be reconciled successfully afterwards. This is achieved by the extension webhook, which adds an annotation shoot.gardener.cloud/tasks= deployInfrastructure when the shoot is changed. 
```

### DIFF
--- a/charts/gardener-extension-admission-alicloud/charts/application/templates/mutatingwebhook-mutator.yaml
+++ b/charts/gardener-extension-admission-alicloud/charts/application/templates/mutatingwebhook-mutator.yaml
@@ -1,0 +1,35 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: {{ include "name" . }}
+webhooks:
+- name: shoots.mutation.alicloud.provider.extensions.gardener.cloud
+  rules:
+  - apiGroups:
+    - core.gardener.cloud
+    apiVersions:
+    - v1alpha1
+    - v1beta1
+    operations:
+    - UPDATE
+    resources:
+    - shoots
+  admissionReviewVersions:
+    - v1
+    - v1beta1
+  clientConfig:
+    {{- if .Values.global.virtualGarden.enabled }}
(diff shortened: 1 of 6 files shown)


EmoinLanyu (dismissed):
Since changing the encryption setting will result in adding new annotation and reconciling the Shoot, I think we also need to document this behavior somewhere.

jia-jerry:
> Since changing the encryption setting will result in adding new annotation and reconciling the Shoot, I think we also need to document this behavior somewhere.

Regarding the auto reconciliation, I mentioned in the user usage doc. Just committed. Regarding the annotation, I just added to the release note as I think it should be transparent to end users.

EmoinLanyu (approved):

PR #339, opened 2021-08-23
Enable ssh key rotation

jia-jerry:

/area control-plane
/kind enhancement
/platform alicloud

For each reconciliation, infra controller will remove all attachment from machine to current used ssh-key.
Fixes https://github.com/gardener/gardener/issues/4395

**Release note**:

```other operator
Gardener can rotate ssh keys for AliCloud now. With this version, you should use gardenctl@v0.24.3 to ssh to AliCloud nodes.
```


charts/gardener-extension-provider-alicloud/values.yaml
```diff
@@ -69,6 +69,7 @@ config:
     storage:
       className: gardener.cloud-fast
       capacity: 25Gi
+  enableSshKeyRotate: false
```
rfranzke, line 72: Why is this configurable in the provider config of the Alicloud extension? I mean, why would I want to disable this? The SSH rotation feature of Gardener is not optional but rather mandatory, so the provider extensions should always support it.

pkg/controller/infrastructure/actuator.go
```diff
@@ -507,6 +544,10 @@ func (a *actuator) reconcile(ctx context.Context, infra *extensionsv1alpha1.Infr
 		return err
 	}
 
+	if err = a.ensureOldSSHKeyDetached(ctx, infra); err != nil {
```
rfranzke, line 547: How will this work - you would always detach the SSH key from the Alicloud machines, right? The key will then only be managed by the `cloud-config-downloader` running on the nodes? If yes, is it possible to not create the machines with the key in the first place, i.e., only use the `cloud-config-downloader` for managing the SSH keys?

jia-jerry: Good suggestion, I will remove ssh-key generation from terraform and do some other follow up changes and test.

pkg/apis/config/v1alpha1/types.go
```diff
@@ -34,11 +34,13 @@ type ControllerConfiguration struct {
 	// settings for the proxy server to use when communicating with the apiserver.
 	// +optional
 	ClientConnection *componentbaseconfigv1alpha1.ClientConnectionConfiguration `json:"clientConnection,omitempty"`
+
```
jia-jerry, line 37:
```suggestion
```

pkg/apis/config/v1alpha1/types.go
```diff
@@ -39,6 +39,7 @@ type ControllerConfiguration struct {
 	MachineImageOwnerSecretRef *corev1.SecretReference `json:"machineImageOwnerSecretRef,omitempty"`
 	// WhitelistedImageIDs specifies an array of image IDs that will bypass image sharing.
 	WhitelistedImageIDs []string `json:"whitelistedImageIDs,omitempty"`
+
```
jia-jerry, line 42:
```suggestion
```

pkg/apis/config/types.go
```diff
@@ -32,11 +32,13 @@ type ControllerConfiguration struct {
 	// ClientConnection specifies the kubeconfig file and client connection
 	// settings for the proxy server to use when communicating with the apiserver.
 	ClientConnection *componentbaseconfig.ClientConnectionConfiguration
+
```
jia-jerry, line 35:
```suggestion
```

pkg/apis/config/types.go
```diff
@@ -37,6 +37,7 @@ type ControllerConfiguration struct {
 	MachineImageOwnerSecretRef *corev1.SecretReference
 	// WhitelistedImageIDs specifies an array of image IDs that will bypass image sharing.
 	WhitelistedImageIDs []string
+
```
jia-jerry, line 40:
```suggestion
```

rfranzke (approved):
/lgtm
/needs second-opinion
/invite @ialidzhikov

pkg/controller/infrastructure/actuator.go
```diff
@@ -59,7 +59,7 @@ var StatusTypeMeta = func() metav1.TypeMeta {
 }()
 
 // NewActuator instantiates an actuator with the default dependencies.
-func NewActuator(machineImageOwnerSecretRef *corev1.SecretReference, whitelistedImageIDs []string) infrastructure.Actuator {
+func NewActuator(machineImageOwnerSecretRef *corev1.SecretReference, whitelistedImageIDs []string, enableSshKeyRotation bool) infrastructure.Actuator {
```
ialidzhikov, line 62: Unused param `enableSshKeyRotation`.

```suggestion
func NewActuator(machineImageOwnerSecretRef *corev1.SecretReference, whitelistedImageIDs []string) infrastructure.Actuator {
```

pkg/controller/infrastructure/add.go
```diff
@@ -38,13 +38,15 @@ type AddOptions struct {
...
+	// EnableSshKeyRotation ensures ssh key rotation always succeeds by detaching all VM instances using the old ssh key.
+	EnableSshKeyRotation bool
 }
 
 // AddToManagerWithOptions adds a controller with the given AddOptions to the given manager.
 // The opts.Reconciler is being set with a newly instantiated actuator.
 func AddToManagerWithOptions(mgr manager.Manager, options AddOptions) error {
 	return infrastructure.Add(mgr, infrastructure.AddArgs{
-		Actuator:          NewActuator(options.MachineImageOwnerSecretRef, options.WhitelistedImageIDs),
+		Actuator:          NewActuator(options.MachineImageOwnerSecretRef, options.WhitelistedImageIDs, options.EnableSshKeyRotation),
```
ialidzhikov, line 49:
```suggestion
		Actuator:          NewActuator(options.MachineImageOwnerSecretRef, options.WhitelistedImageIDs),
```

pkg/controller/infrastructure/add.go
```diff
@@ -38,13 +38,15 @@ type AddOptions struct {
 	MachineImageOwnerSecretRef *corev1.SecretReference
 	// WhitelistedImageIDs specifies an array of image IDs that will bypass image sharing.
 	WhitelistedImageIDs []string
+	// EnableSshKeyRotation ensures ssh key rotation always succeeds by detaching all VM instances using the old ssh key.
+	EnableSshKeyRotation bool
```
ialidzhikov, line 42: Is it needed after https://github.com/gardener/gardener-extension-provider-alicloud/pull/339#discussion_r694072155?

```suggestion
```

pkg/alicloud/client/client.go
```diff
@@ -222,6 +224,60 @@ func (c *ecsClient) ShareImageToAccount(ctx context.Context, regionID, imageID,
 	return err
 }
 
+// DetachECSInstancesFromSshKey finds all ECS instances and detach them from the specified SSH key pair.
+func (c *ecsClient) DetachECSInstancesFromSshKey(keyName string) error {
```
ialidzhikov, line 228: Currently this func is not a trivial one and there are no unit tests related to it. I don't know what kind of possibilities the alicloud sdk has for mocking/faking but it would be nice to have unit tests for this function, if possible.

pkg/apis/alicloud/types_infrastructure.go
```diff
@@ -110,7 +110,9 @@ type NatGatewayConfig struct {
 type InfrastructureStatus struct {
 	metav1.TypeMeta
 
-	VPC         VPCStatus
+	VPC VPCStatus
+
+	// KeyPairName should not be used as ssh-key-pair will not generated by terraform.
```
ialidzhikov, line 115: FYI https://github.com/golang/go/wiki/Deprecated

```suggestion
	// KeyPairName is the key pair name.
	//
	// Deprecated: the field should not be used anymore as the SSH key pair is no longer generated by terraform.
```

pkg/controller/infrastructure/actuator.go
```diff
@@ -294,6 +292,40 @@ func (a *actuator) isWhitelistedImageID(imageID string) bool {
...
+	if infra.Status.ProviderStatus == nil {
+		return nil
+	}
+
+	infrastructureStatus := &alicloudv1alpha1.InfrastructureStatus{}
+	if _, _, err := a.Decoder().Decode(infra.Status.ProviderStatus.Raw, nil, infrastructureStatus); err != nil {
+		return err
+	}
+
+	// nolint
```
ialidzhikov, line 307: What is the reason this `// no lint` comment? `make check` is successful even without it.

pkg/alicloud/client/client.go
ialidzhikov, line 228:
```suggestion
func (c *ecsClient) DetachECSInstancesFromSSHKeyPair(keyName string) error {
```

pkg/controller/infrastructure/actuator.go
```diff
@@ -294,6 +292,40 @@ func (a *actuator) isWhitelistedImageID(imageID string) bool {
 	return false
 }
 
+// ensureOldSSHKeyDetached ensures the compatibility when ssh key is generated in the current cluster via terraform.
+func (a *actuator) ensureOldSSHKeyDetached(ctx context.Context, infra *extensionsv1alpha1.Infrastructure) error {
+
```
ialidzhikov, line 297:
```suggestion
```

pkg/controller/infrastructure/templates/main.tpl.tf
```diff
@@ -4,12 +4,6 @@ provider "alicloud" {
   region = "{{ .alicloud.region }}"
 }
 
-// Import an existing public key to build a alicloud key pair
-resource "alicloud_key_pair" "publickey" {
-  key_pair_name = "{{ .clusterName }}-ssh-publickey"
-  public_key    = "{{ .sshPublicKey }}"
-}
-
```
ialidzhikov, line 12: Okay, the key pair resource is removed from the terraform config but where is the key pair creation logic? Don't we still need to create an ECS key pair?

pkg/controller/infrastructure/actuator.go
```diff
@@ -507,6 +536,10 @@ func (a *actuator) reconcile(ctx context.Context, infra *extensionsv1alpha1.Infr
 		return err
 	}
 
+	if err = a.ensureOldSSHKeyDetached(ctx, infra); err != nil {
```
ialidzhikov, line 539: Why this func is invoked exactly here? I guess the more appropriate place would be before creating the initializer.

```suggestion
	if err = a.ensureOldSSHKeyDetached(ctx, infra); err != nil {
		return err
	}

	tf, err := common.NewTerraformerWithAuth(a.logger, a.terraformerFactory, a.RESTConfig(), TerraformerPurpose, infra)
	if err != nil {
		return err
	}

	initializerValues, err := a.getInitializerValues(ctx, tf, infra, config, credentials)
	if err != nil {
		return err
	}

	initializer, err := a.newInitializer(infra, config, initializerValues, stateInitializer)
	if err != nil {
		return err
	}

	if err := tf.InitializeWith(ctx, initializer).Apply(ctx); err != nil {
		return errors.Wrapf(err, "failed to apply the terraform config")
	}
```

ialidzhikov (changes requested):

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
```diff
@@ -113,8 +113,10 @@ type NatGatewayConfig struct {
 type InfrastructureStatus struct {
 	metav1.TypeMeta `json:",inline"`
 
-	VPC         VPCStatus `json:"vpc"`
-	KeyPairName string    `json:"keyPairName"`
+	VPC VPCStatus `json:"vpc"`
+
+	// KeyPairName should not be used as ssh-key-pair will not generated by terraform.
```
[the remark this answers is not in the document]
jia-jerry: I used to set it as `Deprecated`. But I can't pass the `make check` because the generated files uses this field and I have no chance to mark it as `no-lint`.

pkg/controller/infrastructure/actuator.go
jia-jerry: I originally annotate the KeyPairName to be `deprecated`. But failed in generated file. I will remove this.

pkg/alicloud/client/client.go
jia-jerry: Good suggestion. Unfortunately, aliyunclient only expose a struct type. I would not introduce complicated reflection techniques for to implement this UT.

pkg/controller/infrastructure/templates/main.tpl.tf
jia-jerry: We don't need a SSH key when we create a virtual machine. We rely on https://github.com/gardener/gardener/blob/master/pkg/operation/botanist/component/extensions/operatingsystemconfig/original/components/gardeneruser/templates/scripts/reconcile.tpl.sh to set ssh-key

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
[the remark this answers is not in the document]
ialidzhikov: I see. Then let's mark the field as deprecated and use nolint in the few usages.

```diff
--- a/pkg/controller/infrastructure/actuator.go
+++ b/pkg/controller/infrastructure/actuator.go
@@ -303,8 +303,7 @@ func (a *actuator) ensureOldSSHKeyDetached(ctx context.Context, infra *extension
                return err
        }

-       // nolint
-       if infrastructureStatus.KeyPairName == "" {
+       if infrastructureStatus.KeyPairName == "" { //nolint:staticcheck
                return nil
        }

@@ -318,9 +317,9 @@ func (a *actuator) ensureOldSSHKeyDetached(ctx context.Context, infra *extension
                return err
        }

-       a.logger.V(2).Info("Detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)
-       err = shootAlicloudECSClient.DetachECSInstancesFromSshKey(infrastructureStatus.KeyPairName)
-       a.logger.V(2).Info("Finished detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)
+       a.logger.V(2).Info("Detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)          //nolint:staticcheck
+       err = shootAlicloudECSClient.DetachECSInstancesFromSshKey(infrastructureStatus.KeyPairName)                           //nolint:staticcheck
+       a.logger.V(2).Info("Finished detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName) //nolint:staticcheck

        return err
 }
```

pkg/controller/infrastructure/templates/main.tpl.tf
ialidzhikov: Hm, I see https://github.com/gardener/gardenctl/pull/519. Is it correct that in order to have ssh working for gardenclt@v1 after this change, an operator should use version of gardenctl that includes https://github.com/gardener/gardenctl/pull/519? If this is the case we need to add a release not of type breaking for operators and we should ping for a new version of gardenctl that includes https://github.com/gardener/gardenctl/pull/519.

ialidzhikov: I can confirm that gardenctl@v0.24.2 won't work with this change. gardener@master (hopefully gardenctl@0.25.0) will be required. So let's a release note of type breaking for operators that describes this case and let's ping the gardenctl folks for a release that includes gardener/gardenctl#519.

pkg/controller/infrastructure/templates/main.tpl.tf
```diff
@@ -4,12 +4,6 @@ provider "alicloud" {
   region = "{{ .alicloud.region }}"
 }
 
-// Import an existing public key to build a alicloud key pair
-resource "alicloud_key_pair" "publickey" {
```
ialidzhikov, line 8: I also believe that the infrastructure integration test has to be adapted. Currently the test should have expectations that the ECS key pair exists. I don't see any changes in the integration test as part of this PR. Can you please check?

ialidzhikov:
I would also appreciate if you can squash to single commit at the end - currently the additional commits do not make sense on their own.

pkg/controller/infrastructure/actuator.go
jia-jerry: I don't have any preference. I originally tried to ensureOldSSHKeyDetached before terraformer logic segment instead of terraformer execution. If you have no objection, I would keep it unchanged.

pkg/controller/infrastructure/templates/main.tpl.tf
jia-jerry: Yes. will check it.

pkg/controller/infrastructure/templates/main.tpl.tf
jia-jerry: Fix  gardener/gardenctl#519 is not target for ssh-key rotation. It fixed crash issue because of a regression. The  current master branch can ssh to both old and future(ssh key rotation) version of AliCloud extension.

ialidzhikov: Yeah, this is correct. But currently latest gardenctl (0.24.2) won't work against provider extension that includes this PR.

jia-jerry: Hmm, I made a mistake. gardenctl has the regression in master not in v0.24.2.  I will release a note in this PR.

jia-jerry:
@ialidzhikov  Thank you for your suggestion. Would you mind to give another review?

pkg/controller/infrastructure/templates/main.tpl.tf
ialidzhikov: I cannot follow what you mean by your latest comment. Yesterday I checked locally my statement 

> Yeah, this is correct. But currently latest gardenctl (0.24.2) won't work against provider extension that includes this PR. That why I wrote that first gardenctl@0.25.0 is required before considering this PR.

and made sure that it is the case. So I think we need gardenctl release that includes gardener/gardenctl#519, before considering this PR.

pkg/apis/alicloud/types_infrastructure.go
ialidzhikov: I don't think this comment is resolved.

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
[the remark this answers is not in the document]
ialidzhikov: I don't think this comment is resolved.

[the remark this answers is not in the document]
jia-jerry: I can't do it for `generated` files. I will remove it in future..

pkg/controller/infrastructure/actuator.go
ialidzhikov: I think the better approach is the one proposed by me. The tf init preparations should happen before the Apply operation. Currently there is an operation (`ensureOldSSHKeyDetached`) between the tf init preparations and the Apply itself.

ialidzhikov (commented):
/hold
until the backwards compatibility concerns related to gardenctl are resolved/clarified

pkg/apis/alicloud/v1alpha1/types_infrastructure.go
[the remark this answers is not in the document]
ialidzhikov: > I can't do it for generated files. I will remove it in future..

Did you checked this locally?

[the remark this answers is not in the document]
ialidzhikov: For me, `make check` is passing with the following diff:

```diff
diff --git a/pkg/apis/alicloud/types_infrastructure.go b/pkg/apis/alicloud/types_infrastructure.go
index ef83404b..e1e89fe3 100644
--- a/pkg/apis/alicloud/types_infrastructure.go
+++ b/pkg/apis/alicloud/types_infrastructure.go
@@ -112,7 +112,9 @@ type InfrastructureStatus struct {
 
 	VPC VPCStatus
 
-	// KeyPairName should not be used as ssh-key-pair will not generated by terraform.
+	// KeyPairName is the key pair name.
+	//
+	// Deprecated: the field should not be used anymore as the SSH key pair is no longer generated by terraform.
 	KeyPairName string
 
 	// MachineImages is a list of machine images that have been used in this infrastructure. Usually, the extension controller
diff --git a/pkg/apis/alicloud/v1alpha1/types_infrastructure.go b/pkg/apis/alicloud/v1alpha1/types_infrastructure.go
index 71920833..2a099ebc 100644
--- a/pkg/apis/alicloud/v1alpha1/types_infrastructure.go
+++ b/pkg/apis/alicloud/v1alpha1/types_infrastructure.go
@@ -115,7 +115,9 @@ type InfrastructureStatus struct {
 
 	VPC VPCStatus `json:"vpc"`
 
-	// KeyPairName should not be used as ssh-key-pair will not generated by terraform.
+	// KeyPairName is the key pair name.
+	//
+	// Deprecated: the field should not be used anymore as the SSH key pair is no longer generated by terraform.
 	KeyPairName string `json:"keyPairName"`
 
 	// MachineImages is a list of machine images that have been used in this infrastructure. Usually, the extension controller
diff --git a/pkg/controller/infrastructure/actuator.go b/pkg/controller/infrastructure/actuator.go
index 9ddc4543..38acbe66 100644
--- a/pkg/controller/infrastructure/actuator.go
+++ b/pkg/controller/infrastructure/actuator.go
@@ -303,7 +303,7 @@ func (a *actuator) ensureOldSSHKeyDetached(ctx context.Context, infra *extension
 		return err
 	}
 
-	if infrastructureStatus.KeyPairName == "" {
+	if infrastructureStatus.KeyPairName == "" { //nolint:staticcheck
 		return nil
 	}
 
@@ -317,9 +317,9 @@ func (a *actuator) ensureOldSSHKeyDetached(ctx context.Context, infra *extension
 		return err
 	}
 
-	a.logger.V(2).Info("Detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)
-	err = shootAlicloudECSClient.DetachECSInstancesFromSSHKeyPair(infrastructureStatus.KeyPairName)
-	a.logger.V(2).Info("Finished detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)
+	a.logger.V(2).Info("Detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)          //nolint:staticcheck
+	err = shootAlicloudECSClient.DetachECSInstancesFromSSHKeyPair(infrastructureStatus.KeyPairName)                       //nolint:staticcheck
+	a.logger.V(2).Info("Finished detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName) //nolint:staticcheck
 
 	return err
 }
```

pkg/controller/infrastructure/templates/main.tpl.tf
jia-jerry: I asked @neo-liang-sap  to  do a release.

ialidzhikov:
/unhold
as there is now https://github.com/gardener/gardenctl/releases/tag/v0.24.3

pkg/controller/infrastructure/actuator.go
jia-jerry: I will take your advise. I messed your proposal with my real implementation from this change proposal.  sorry~

pkg/apis/alicloud/types_infrastructure.go
```diff
@@ -110,7 +110,9 @@ type NatGatewayConfig struct {
 type InfrastructureStatus struct {
 	metav1.TypeMeta
 
-	VPC         VPCStatus
+	VPC VPCStatus
+
+	// Deprecated:  KeyPairName should not be used as ssh-key-pair will not generated by terraform.
```
ialidzhikov, line 115: Hm, I see. Then let's add again similar rule to `.golangci.yaml`. This should do the trick.

```diff
diff --git a/.golangci.yaml b/.golangci.yaml
index d54e5358..8bb89792 100644
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,3 +5,9 @@ run:
 linters:
   disable:
   - unused
+
+issues:
+  exclude-rules:
+  - linters:
+    - staticcheck
+    text: "SA1019:" # Excludes messages where deprecated variables are used
```

ialidzhikov, line 115: Btw, my initial suggestion was a different one:

```suggestion
	// KeyPairName is the key pair name.
	//
	// Deprecated: the field should not be used anymore as the SSH key pair is no longer generated by terraform.
```

jia-jerry: I also constraint some files.

pkg/controller/infrastructure/actuator.go
```diff
@@ -294,6 +292,41 @@ func (a *actuator) isWhitelistedImageID(imageID string) bool {
...
+
+	shootAlicloudECSClient, err := a.newClientFactory.NewECSClient(infra.Spec.Region, shootCloudProviderCredentials.AccessKeyID, shootCloudProviderCredentials.AccessKeySecret)
+	if err != nil {
+		return err
+	}
+	// nolint
+	a.logger.V(2).Info("Detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)
+	// nolint
+	err = shootAlicloudECSClient.DetachECSInstancesFromSSHKeyPair(infrastructureStatus.KeyPairName)
+	// nolint:static
```
ialidzhikov, line 324: I think we better ignore this file as well in `.golangci.yaml` and remove the `nolint`s in this func.

Locally I get:

```
$ make check
> Check
Executing golangci-lint
WARN [runner/nolint] Found unknown linters in //nolint directives: static
pkg/controller/infrastructure/actuator.go:325:86: SA1019: infrastructureStatus.KeyPairName is deprecated: the field should not be used anymore as the SSH key pair is no longer generated by terraform. (staticcheck)
	a.logger.V(2).Info("Finished detaching ssh key pair from ECS instances", "keypair", infrastructureStatus.KeyPairName)
	                                                                                    ^
make: *** [check] Error 1
```

I don't know why the verify step does not fail with the same.

test/integration/infrastructure/common.go
```diff
@@ -24,12 +24,9 @@ const (
...
 	natGatewayType = "Enhanced"
 
-	secretName = "cloudprovider"
-
-	availableStatus = "Available"
-
-	sshPublicKey       = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDSb1DJfupnWTfKJ0fmRGgnSx8A2/pRd5oC49qE1jFX+/J9L01jUyLc5sBKZXVkfU5q5h0JfbkhJXSIkzqE+rNPnJBI4e+8Lo2TVWLAvVZRA9Fg9Dk3mgkdVB+9qW2mIqtJF5GOWKuk7HkObwpY1pX8kHC/LJVfpNQpVBqWef0WJj6vbyjhlZ3vgRxK9I6wdJzjUYtNsDhvvBTy/IBg/xp82w9T2r3GVfnaTLMeQCW9mPviDKnQsrWMgVb2A0Z4c62EbzzLzQV4ScVJ6JMgOgkMqEPdbnKF8dEQcSu+/DQZoZt56Aeov7T4oamahj9/rIDX+WR1nOcfntIdhCyoB4lISkNFz/MlPC7O8HwJk4P7rojLGNk6xmn6NxY5CJGC2dVxFsb1bmm+fKHAp62mgwEoFZcDyIkcsmnmnID9u0rJNyMz84YUGZ/jEz8LePujDHcXiqgoLsKJ8gNRneISL9+m9s1VK7WxDDIbq8iWzR7XfAVE/GzKpVYkqrWCvjKEeFIDuDUnf3jghQCQMsXnJM7zGWr1tl+Dvl2Avxmj2xyUJXYHbXbl2aM434DgQySnV8JPzYH7EsTmvuhdb8SJIbb/NonFsSM+72HpSzVc083x4B++VL7oP1X8cly62pFVM1fi8sxBio48Hq5SmAUu9T4wUY4J+AKU6osFA/ATlMCIiQ== your_email@example.com"
-	sshPublicKeyDigest = "b9b39384513d9300374c98ccc8818a8b"
+	secretName           = "cloudprovider"
+	enableEncryptedImage = false
```
ialidzhikov, line 28: Wrt to commit 4fb0594119aea08dbe7be3113542ed5ca339626e, do we disable the encrypted image testing with this commit? If yes, what is the reason for that? In general disabled/skipped test equals no test -> no way to catch any potential issue/regression. If we have issues/flakes with this test shouldn't we follow-up in separate issue/PR?

ialidzhikov (commented):
Generally looks good. Minor nit and a question inline.
Can I also ask you to squash the commits related to the ssh key pair removal? Another good candidate for a separate commit is a one that disables the encrypted image testing (but I added a question/concern about it).

test/integration/infrastructure/common.go
jia-jerry: It takes too long to encrypt an image. There is an bug in AliCloud. Even the encrypted image is deleted, there is OSS file still remaining. With time, there will be much leftovers.

And also, the encrypted image is covered by integration test. I would suggest to disable it for now.

pkg/controller/infrastructure/actuator.go
jia-jerry: I will remove `static`. The purpose for Deprecated is suggest people not to use it. If we ignore this file, and people may also use this field freely. This is not desired.

test/integration/infrastructure/common.go
ialidzhikov: > And also, the encrypted image is covered by integration test.

The test you mention here is a test executed by testmachinery against existing Gardener landscapes. I would say that both tests are different. We can execute the current test against a PR and in this way catch any potential issues on PR level. The other test is executed against a Gardener landscape and it is usually executed against released version of provider-alicloud. So the fact that there is a testmachinery test for the scenario does not allow us to drop/disable the current envtest.

> It takes too long to encrypt an image. There is an bug in AliCloud. Even the encrypted image is deleted, there is OSS file still remaining. With time, there will be much leftovers.

If there are issues with the test, then please create an issue on provider-alicloud where you properly describe the problems, the actions you took for resolution so far and action items that are needed for enabling the test again. This should be the preferred approach.

jia-jerry: We reported to Ali regarding this issue. OK, take your suggestion we enable it.

ialidzhikov (approved):
/lgtm

PR #383, opened 2021-11-11
enable env DISK_AD_CONTROLLER for csi plugin

jia-jerry:

/area control-plane
/kind enhancement
/platform alicloud

Add DISK_AD_CONTROLLER env to csi plugin. It fix the issue reported [here](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/issues/544), according to AliCloud folks.

Fixes # 

If will fix volume mount failure when several volumes are mounted on one node concurrently.

**Release note**:

```other user
Volume mount failure issue is solved. This issue only happens several volumes are mounted on one node concurrently.
```


EmoinLanyu (approved):
/lgtm

PR #384, opened 2021-11-11
Enable specifying a domain id in `DNSRecord` `spec.zone`

stoyanr:

/area control-plane
/kind enhancement
/platform alicloud

Enables specifying a domain id, in addition to a domain name, in the `spec.zone` field of a `DNSRecord` resource. This is needed to ensure that explicitly configured zones in landscapes and shoot `spec.dns.providers` configurations work for both the `DNSRecord` controller and [external-dns-management](https://github.com/gardener/external-dns-management), which uses the domain id as the zone on Alicloud.

Fixes #

* Specifying the domain name, and not the domain id, is still recommended if possible, since the Alicloud API uses the domain name, not the domain id. If the domain id is specified, the domain name has to be figured out at the price of additional API requests that could otherwise be avoided.
* Nevertheless, in order to avoid "overlapping zones" errors from `external-dns-management`, it is important to be able to configure zones explicitly in a way that works for both `external-dns-management` and the `DNSRecord` controller. Since the handling of zones in `external-dns-management` is harder to change, this PR proposes adapting the `DNSRecord` controller to the zone identification used by `external-dns-management`.

**Release note**:

```feature operator
It is now possible to specify a domain id, in addition to a domain name, in the `spec.zone` field of a `DNSRecord` resource.
```

### DIFF
--- a/pkg/alicloud/client/client_dns.go
+++ b/pkg/alicloud/client/client_dns.go
@@ -40,9 +40,10 @@ func (f *clientFactory) NewDNSClient(region, accessKeyID, accessKeySecret string
 	}
 
 	return &dnsClient{
-		Client:       *client,
-		accessKeyID:  accessKeyID,
-		domainsCache: f.domainsCache,
+		Client:            *client,
+		accessKeyID:       accessKeyID,
+		domainsCache:      f.domainsCache,
+		domainsCacheMutex: &f.domainsCacheMutex,
 	}, nil
 }
 
@@ -132,6 +133,15 @@ func (d *dnsClient) DeleteDomainRecords(ctx context.Context, domainName, name, r
 }
 
 func (d *dnsClient) getDomainsWithCache() (map[string]alidns.Domain, error) {
+	// cache.Expiring Get and Set methods are concurrency-safe.
+	// However, if an accessKeyID is not present in the cache and multiple DNSRecords are reconciled at the same time,
+	// it may happen that getDomains is called multiple times instead of just one, so use a mutex to guard against this.
+	// It is ok to use a shared mutex here as far as the number of accessKeyIDs using custom domains is low.
+	// This may need to be revisited with a larger number of such accessKeyIDs to avoid them blocking each other
+	// during the (potentially long-running) call to getDomains.
+	d.domainsCacheMutex.Lock()
+	defer d.domainsCacheMutex.Unlock()
+
 	if v, ok := d.domainsCache.Get(d.accessKeyID); ok {
 		return v.(map[string]alidns.Domain), nil
 	}
--- a/pkg/alicloud/client/client_dns.go
+++ b/pkg/alicloud/client/client_dns.go
@@ -39,27 +39,30 @@ func (f *clientFactory) NewDNSClient(region, accessKeyID, accessKeySecret string
 	}, nil
 }
 
-// GetDomainNames returns a list of all domain names.
-func (d *dnsClient) GetDomainNames(ctx context.Context) ([]string, error) {
-	var domains []string
-	pageSize, pageNumber := 20, 1
-	req := alidns.CreateDescribeDomainsRequest()
-	req.PageSize = requests.NewInteger(pageSize)
-	for {
-		req.PageNumber = requests.NewInteger(pageNumber)
-		resp, err := d.Client.DescribeDomains(req)
-		if err != nil {
-			return nil, err
-		}
-		for _, domain := range resp.Domains.Domain {
-			domains = append(domains, domain.DomainName)
-		}
-		if resp.PageNumber*int64(pageSize) >= resp.TotalCount {
-			break
-		}
-		pageNumber++
+// GetDomainNames returns a map of all domain names mapped to their composite domain names.
+func (d *dnsClient) GetDomainNames(ctx context.Context) (map[string]string, error) {
+	domains, err := d.getDomains()
+	if err != nil {
+		return nil, err
 	}
-	return domains, nil
+	domainNames := make(map[string]string)
+	for _, domain := range domains {
+		domainNames[domain.DomainName] = CompositeDomainName(domain.DomainName, domain.DomainId)
+	}
+	return domainNames, nil
+}
+
+// GetDomainName returns the composite domain name of the domain with the given domain id.
+func (d *dnsClient) GetDomainName(ctx context.Context, domainId string) (string, error) {
+	domains, err := d.getDomains()
+	if err != nil {
+		return "", err
+	}
+	domain, ok := domains[domainId]
+	if !ok {
+		return "", fmt.Errorf("DNS domain with id %s not found", domainId)
+	}
+	return CompositeDomainName(domain.DomainName, domain.DomainId), nil
 }
 
 // CreateOrUpdateDomainRecords creates or updates the domain records with the given domain name, name, record type,
@@ -68,6 +71,7 @@ func (d *dnsClient) GetDomainNames(ctx context.Context) ([]string, error) {
 // * For each element in values that doesn't have an existing domain record, a new domain record is created.
 // * For each existing domain record that doesn't have a corresponding element in values, the existing record is deleted.
 func (d *dnsClient) CreateOrUpdateDomainRecords(ctx context.Context, domainName, name, recordType string, values []string, ttl int64) error {
+	domainName, _ = DomainNameAndId(domainName)
 	rr, err := getRR(name, domainName)
 	if err != nil {
 		return err
@@ -103,6 +107,7 @@ func (d *dnsClient) CreateOrUpdateDomainRecords(ctx context.Context, domainName,
 
 // DeleteDomainRecords deletes the domain records with the given domain name, name and record type.
 func (d *dnsClient) DeleteDomainRecords(ctx context.Context, domainName, name, recordType string) error {
+	domainName, _ = DomainNameAndId(domainName)
 	rr, err := getRR(name, domainName)
 	if err != nil {
 		return err
@@ -117,6 +122,29 @@ func (d *dnsClient) DeleteDomainRecords(ctx context.Context, domainName, name, r
 		}
 	}
 	return nil
+}
+
+// getDomains returns all domains.
+func (d *dnsClient) getDomains() (map[string]alidns.Domain, error) {
+	domains := make(map[string]alidns.Domain)
+	pageSize, pageNumber := 20, 1
+	req := alidns.CreateDescribeDomainsRequest()
+	req.PageSize = requests.NewInteger(pageSize)
+	for {
+		req.PageNumber = requests.NewInteger(pageNumber)
+		resp, err := d.Client.DescribeDomains(req)
+		if err != nil {
+			return nil, err
+		}
+		for _, domain := range resp.Domains.Domain {
+			domains[domain.DomainId] = domain
+		}
+		if resp.PageNumber*int64(pageSize) >= resp.TotalCount {
+			break
+		}
+		pageNumber++
+	}
+	return domains, nil
 }
 
 // getDomainRecords returns the domain records with the given domain name, rr, and record type.
@@ -194,4 +222,22 @@ func isDomainRecordDoesNotExistError(err error) bool {
 		}
 	}
 	return false
+}
+
+// CompositeDomainName composes and returns a composite domain name from the given domain name and id,
+// in the format <domainName>:<domainId>
+func CompositeDomainName(domainName, domainId string) string {
+	if domainId != "" {
+		return domainName + ":" + domainId
+	}
+	return domainName
+}
+
+// DomainNameAndId decomposes the given composite domain name in the format <domainName>:<domainId>
+// into its constituent domain name and id.
+func DomainNameAndId(compositeDomainName string) (string, string) {
+	if parts := strings.Split(compositeDomainName, ":"); len(parts) == 2 {
+		return parts[0], parts[1]
+	}
+	return compositeDomainName, ""
 }
--- a/pkg/alicloud/client/client_dns.go
+++ b/pkg/alicloud/client/client_dns.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/gardener/gardener-extension-provider-alicloud/pkg/alicloud"
 
@@ -25,6 +26,10 @@ import (
 	"github.com/aliyun/alibaba-cloud-sdk-go/sdk/requests"
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/alidns"
 	"github.com/gardener/gardener/pkg/utils"
+)
+
+const (
+	domainsCacheTTL = 1 * time.Hour
 )
 
 // NewDNSClient creates a new DNS client with given region, accessKeyID, and accessKeySecret.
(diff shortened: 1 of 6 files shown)


pkg/controller/dnsrecord/actuator.go
```diff
@@ -144,8 +145,21 @@ func (a *actuator) Migrate(ctx context.Context, dns *extensionsv1alpha1.DNSRecor
...
 	switch {
-	case dns.Spec.Zone != nil && *dns.Spec.Zone != "":
-		return *dns.Spec.Zone, nil
+	case dns.Spec.Zone != nil && *dns.Spec.Zone != "" && (dns.Status.Zone == nil || *dns.Status.Zone == "" || !zoneMatchesDomainName(*dns.Spec.Zone, *dns.Status.Zone)):
+		if isDomainName(*dns.Spec.Zone) {
+			return *dns.Spec.Zone, nil
+		}
+		// The value specified in dns.Spec.Zone is not a domain name, so assume it's a domain id,
+		// and try to determine the domain name by getting the name of the domain with this id
+		domainName, err := dnsClient.GetDomainName(ctx, *dns.Spec.Zone)
```
MartinWeindel, line 154: I have concerns about the case if the value specified in the dns.Spec.Zone is a domain id.
In this case you have a `DescribeDomains` request for the first reconciliation of every single `DNSRecord`.
Wouldn't it be better if the result is cached? The domains will only change very infrequently. And if I count correctly there can be four `DNSRecords` for a cluster.

MartinWeindel (commented):
Looks good to me, but I have one comment regarding caching of the domains.

pkg/controller/dnsrecord/actuator.go
stoyanr: Yes, you are right, I deliberately wanted to avoid caching in `DNSRecord` controllers to keep them simple, but I guess in this case I should make an exception, since the main mechanism to avoid getting the domains (namely specifying them explicitly) is essentially broken if domain ids, and not domain names, are specified.

MartinWeindel (approved):
/lgtm

pkg/controller/dnsrecord/actuator.go
stoyanr: OK, I added caching for domains now. It uses an expiring cache with a TTL of 1 hour, so `DescribeDomains` results will be cached (per `accessKeyId`) for 1 hour before they expire and have to be fetched again. Do you consider this good enough, or would you propose a different caching strategy or TTL value?

jia-jerry: @MartinWeindel  I don't remember if you implement DNS client rate limit before. In future, I would also suggest cache client too and enforce ratelimit config for each client.

MartinWeindel: @jia-jerry Yes we use a client-side rate limiter for all providers in external-dns-management. The [TokenBucketRateLimiter](https://github.com/kubernetes/client-go/blob/master/util/flowcontrol/throttle.go#L63) is used for this purpose.

pkg/alicloud/client/client_dns.go
```diff
@@ -35,31 +40,36 @@ func (f *clientFactory) NewDNSClient(region, accessKeyID, accessKeySecret string
...
+func (d *dnsClient) GetDomainName(ctx context.Context, domainId string) (string, error) {
+	domains, err := d.getDomainsWithCache()
+	if err != nil {
+		return "", err
+	}
+	domain, ok := domains[domainId]
+	if !ok {
+		return "", fmt.Errorf("DNS domain with id %s not found", domainId)
+	}
+	return CompositeDomainName(domain.DomainName, domain.DomainId), nil
```
jia-jerry, line 72: I would suggest to deprecate `zone` in status, and introduce 2 fields `zoneId` and `zoneName`.  Using `compositeDomainName` is a little tricky.

pkg/controller/dnsrecord/actuator.go
```diff
@@ -144,8 +145,21 @@ func (a *actuator) Migrate(ctx context.Context, dns *extensionsv1alpha1.DNSRecor
...
+		if err != nil {
+			return "", &controllererror.RequeueAfterError{
+				Cause:        fmt.Errorf("could not get DNS domain name for domain id %s: %+v", *dns.Spec.Zone, err),
+				RequeueAfter: requeueAfterOnProviderError,
+			}
+		}
+		a.logger.Info("Got DNS domain name", "domainName", domainName, "dnsrecord", kutil.ObjectName(dns))
+		return domainName, nil
 	case dns.Status.Zone != nil && *dns.Status.Zone != "":
```
jia-jerry, line 163: How about when a dns zone is switched to another account?
If `dns.Spec.Zone` is not specified, it seems we always go to `default` case.

jia-jerry (changes requested):

pkg/alicloud/client/client_dns.go
stoyanr: This would be really hard to do. The `DNSRecord` resource is used in g/g and also across 5 different providers, and alicloud is the only one where we would use both `zoneId` and `zoneName` (and only because historically `external-dns-management` uses domain ids that can't be used directly with the Alicloud API). I would like to contain this alicloud-specific issue to alicloud, and avoid changing the API and all other providers just because of it.

pkg/controller/dnsrecord/actuator.go
stoyanr: The `default` case will query the domains (via `DescribeDomains`) and will try to determine the correct domain name based on the dnsrecord name (using the "longest matching" approach implemented in `dnsrecord.FindZoneForName`). There is no change here. What is the actual issue that you see? Perhaps I don't actually understand what you mean by "a dns zone is switched to another account".

pkg/controller/dnsrecord/actuator.go
stoyanr: Such a rate limiter was recently added to the AWS provider extension, see gardener/gardener-extension-provider-aws/pull/425, after we experienced actual issues with exceeding the route53 rate limits on some of our landscapes. We could something similar here as well (in a different PR). What are the Alicloud rate limits? Do you expect that they could be exceeded, have we actually experienced such issues in the past?

pkg/controller/dnsrecord/actuator.go
jia-jerry: It is a corner case. Please just ignore it in this PR.
Here is example:
- domain `example.com` is in AliCloud *account-a*
- DNS record is successfully reconciled, with status maintained correctly.
- domain `example.com` is designated in AliCloud *account-b*

It seems that the extension can't handle this case.

pkg/alicloud/client/client_dns.go
jia-jerry: How about we still only store `domain name` in `zone` field? It seems no side effect. 
It could be simpler and easy to understand in the code.

The purpose of this PR is to handle *union typed* value for `spec.zone`. I think the only place to consider it is in  function `getDomainName`.

Maybe you have concerns about an extra api call to alicloud. But I don't think it is an issue. 
- We already cached it.
- We will later enforce client ratelimit.

You know, we can't avoid AliCloud API call because of the following method `CreateOrUpdateDomainRecords`.

pkg/controller/dnsrecord/actuator.go
jia-jerry: We encountered it before https://github.com/gardener/external-dns-management/pull/77.

jia-jerry: @stoyanr, would you please take care of https://github.com/gardener/gardener-extension-provider-alicloud/issues/387?

shaoyongfeng (commented):
the pr itself looks good to me. 
But it's more important for me to document how to specify zones in landscape.yml, so gardener ops knows how to config it. 
Since dnsrecord controller can handle both domain name and ID, @MartinWeindel  it's good that you can provide such docu from extern dns management point of view.

pkg/alicloud/client/client_dns.go
stoyanr: > How about we still only store domain name in zone field? It seems no side effect.

The reason I would like to have the domain id as well is to make sure that `DescribeDomains` will be called only if the domain id in `spec.zone` changed (or was added where it was previously missing), or if it's the very first reconciliation and the status is not updated yet. I would not really like to perform a `DescribeDomains` on every call if it can be avoided, no matter that we cache the result. And introducing a client side rate limiter is no real solution either (you will still get errors if the configured rate limit is exceeded, they tend to resolve faster than if you exceed the provider rate limit but they are still very visible and annoying).

> You know, we can't avoid AliCloud API call because of the following method CreateOrUpdateDomainRecords.

This has no impact on this discussion or this PR, it's really only about avoiding redundant calls to `DescribeDomains`.

pkg/controller/dnsrecord/actuator.go
stoyanr: If the domain is no longer accessible from the account, this would result in an error, with or without this PR. Since this PR introduced caching, `GetDomainName` might return an outdated result (still containing the domain), and so the error won't be `could not find DNS domain name for name ...`, but a different error when trying to manipulate an actual record. But in essence, I don't think this PR changes what happens in this case in a way that we would need to worry about.

Was this what you were concerned about, or perhaps I am still missing your point?

jia-jerry: Never mind, it is a corner case and it has no relationship with PR. We could discuss it if some issue happens.

pkg/alicloud/client/client_dns.go
```diff
@@ -119,6 +131,41 @@ func (d *dnsClient) DeleteDomainRecords(ctx context.Context, domainName, name, r
...
 
+func (d *dnsClient) getDomainsWithCache() (map[string]alidns.Domain, error) {
+	if v, ok := d.domainsCache.Get(d.accessKeyID); ok {
+		return v.(map[string]alidns.Domain), nil
+	}
+	domains, err := d.getDomains()
+	if err != nil {
+		return nil, err
+	}
+	d.domainsCache.Set(d.accessKeyID, domains, domainsCacheTTL)
```
jia-jerry, line 142: It still has issue when a lot of dns records  with the same accessKeyId are reconciled. This can happen because `d.getDomains()` takes long time enough. As a result, lots of `dnsClient` will call `d.getDomains()` inevitably though most of them are expected to go through the cache.

I would suggest introduce extra locks:
1, maintain a globla lock table accessKeyIDLockTable map[accessKeyID][Lock]
2, a global lock: gCacheLock
3. Check domain cache with accessKeyID, if it doesn't exist, go to step 4.
4. retreive lock from accessKeyIDLockTable, if doesn't exist,go to 5. Otherwise, goto 6. The lock for accesKeyID level is accessKeyIDLock.
5. gCacheLock.Lock(),  put a lock in accessKeyIDLock  for the specified accessKeyID, gCacheLock.UnLock(). 
6. accessKeyIDLock.lock and fill with domain cache.

pkg/alicloud/client/client_dns.go
jia-jerry: I think if we maintain the domain cache correctly, it doens't have any issue.

pkg/controller/dnsrecord/actuator.go
stoyanr: OK, resolving this discussion then, let me know if you think it should be addressed in a different PR.

pkg/alicloud/client/client_dns.go
stoyanr: In my opinion, caching comes with its own set of issues due to the cache potentially becoming outdated (even if for an hour, this can still cause friction). I introduced it there because I don't really know what else to propose to reduce the number of redundant `DescribeDomains` calls, I just hope (but I am not certain) we will not have to revisit it. I still strongly suggest to keep the logic that relies on the `zone` field in status. This is what all other `DNSRecord` controllers do and I see no reason to abandon it here.

What is your concern actually? Having the `zone` field in the format `<domain-name>:<domain-id>` is not ideal, but certainly not a big issue either?

pkg/alicloud/client/client_dns.go
stoyanr: Thanks for spotting this, you are right this can happen, and indeed we need additional locking to prevent it. But I don't think we need so many locks - a single lock on the (shared) cache should be enough, to ensure only a single `d.getDomains()` is performed. `cache.Expring` itself is already concurrency safe, we don't need to do anything special to protect its entries.

pkg/controller/dnsrecord/actuator.go
stoyanr: I will, in a separate PR, as mentioned.

pkg/alicloud/client/client_dns.go
stoyanr: Done, PTAL.

pkg/alicloud/client/client_dns.go
```diff
@@ -119,6 +132,47 @@ func (d *dnsClient) DeleteDomainRecords(ctx context.Context, domainName, name, r
 	return nil
 }
 
+func (d *dnsClient) getDomainsWithCache() (map[string]alidns.Domain, error) {
+	// cache.Expiring Get and Set methods are concurrency-safe
+	// However, if an accessKeyID is not present in the cache and multiple DNSRecords are reconciled at the same time,
+	// it may happen that getDomains is called multiple times instead of just one, so use a mutex to guard against this
+	d.domainsCacheMutex.Lock()
```
jia-jerry, line 139: This is a global lock. Not sure in the normal case, how many accessKeyIDs will be served for a single extension.
It is not a good practice as  `	domains, err := d.getDomains()` is a long call.

stoyanr: In the worst case, as many accessKeyIDs as we have shoots on a single seed (assuming every shoot uses a different accessKeyID, which is very unlikely), and the max number of such shoots is 250 currently (can be configured, but this is the default value, and for many reasons it should not be exceeded by much). In reality, we should assume no more than ~100 accessKeyIDs. But getting all of them to contend for the same lock is very unlikely, since we would need to have `DNSRecords` resources of many different shoots being reconciled at the same time. There are precautions to prevent too many shoots from being reconciled at once (e.g. configurable jitter in `gardenlet`), so I would expect `DSNRecords` of no more than 10-20 different shoots being reconciled at the same time, and it's likely that at least for some of these the domains will be already present in the cache.

I get your point now that it's possible to avoid this by introducing somewhat more complex locking, but I would rather not make it more complex than needed for now.

pkg/alicloud/client/client_dns.go
jia-jerry: It is OK. Just personal preference.

pkg/alicloud/client/client_dns.go
jia-jerry: I also checked current shoots for AliCloud - only one custom domain exists. Current code is OK. Would you please add some comment for this lock to depict the potential issue if more custom domains emerge than expected?

jia-jerry (dismissed):

pkg/alicloud/client/client_dns.go
stoyanr: OK, I added such a comment.

pkg/alicloud/client/client_dns.go
stoyanr: OK, let's keep it as it is then, thanks!

jia-jerry (approved):

PR #405, opened 2021-12-21
🛡 Enable `ServiceAccount` token projection and token requestor

rfranzke:

/area security
/kind enhancement
/merge squash

* When the extension is running on a seed with a gardenlet of at least `v1.37` then
  * `ServiceAccount` token projection is enabled.
  * the components deployed into shoot namespaces
    * uses `ServiceAccount` token projection.
    * uses a token managed by GRM's token requestor instead of a client certificate.

Part of gardener/gardener#4659
Part of gardener/gardener#4878

✅ ~Depends on #403, hence, PR is in draft state.~
✅ ~Depends on #422, hence, PR is in draft state.~

**Release note**:

```noteworthy operator
The extension controller uses a projected `ServiceAccount` token in case it runs on a seed with a gardenlet of at least `v1.37` or higher. Similarly, the components deployed into shoot namespaces will no longer use a client certificate but an auto-rotated `ServiceAccount` token which is only valid for `12h`.
```


charts/internal/seed-controlplane/charts/alicloud-cloud-controller-manager/templates/deployment.yaml
```diff
@@ -33,13 +33,18 @@ spec:
 {{ toYaml .Values.podLabels | indent 8 }}
 {{- end }}
     spec:
+      automountServiceAccountToken: false
```
jia-jerry, line 36: Does it need `security.gardener.cloud/trigger` annotaion?

rfranzke: No, this setting here is only for preventing mounting the default service account into the CCM pod. It uses a kubeconfig (`--kubeconfig`) to talk to the shoot cluster, hence, it does not need a service account in the seed.

jia-jerry: Hope I didn't misled you. Maybe I should not ask you question under this statement. I meant for other controllers like CSI and MCM, they have `security.gardener.cloud/trigger` annotation. How about CCM?

rfranzke: This is not needed because CCM does not use a `ServiceAccount` of the seed cluster.

jia-jerry: I understood. Thanks for explanation.

jia-jerry (approved):
/LGTM

BTW. It took me more than 1 day to review this PR cause I need to go through the related code of resource manger and extension shared components. 

Need a second opinion from a member of Gardener core team.

rfranzke:
Thanks a lot @jia-jerry for spending the time to review this thoroughly! These PR are all "the same" for all provider extensions. Given that @MartinWeindel has performed another thorough review for https://github.com/gardener/gardener-extension-provider-vsphere/pull/192, can we proceed?

jia-jerry:
Thanks for your great enhancement. Brilliant idea! I will merge it.

PR #408, opened 2021-12-22
remove label from machine class if value is empty

shaoyongfeng:
/area control-plane
/kind bug
/platform alicloud

Ali doesn't accept empty value as tag value, but some gardener users such as hanacloud set empty value as label value in their shoot yaml. So we need to filter out such label
Fixes #

**Release note**:
```bug user
Label without value won't be added as tag in ECS on alicloud
```


jia-jerry (dismissed):

jia-jerry (approved):
/lgtm

PR #410, opened 2021-12-23
add region to image encryption ROS template name

jia-jerry:

/area security
/kind bug
/platform alicloud

Add region id to the name of Stack, which is used to encrypt a VM image. It fixes an issue when one AliCloud account is used to create shoot in 2+ regions. 

Reproduce steps:
1. Create a shoot in one region, whose system disk is encrypted. It will be success.
2. Create and shoot in another region, whose system disk is encrypted. It will fail.

Fixes #

**Release note**:

```bugfix user
Now it is possible to create shoots with encrypted system disk in more than one regions.
```


EmoinLanyu (approved):
/lgtm

PR #416, opened 2022-01-07
fix image name dup issue

jia-jerry:

/area control-plane
/kind bug
/platform alicloud

Fixes issue introduced in https://github.com/gardener/gardener-extension-provider-alicloud/pull/410. The image name should be different if the stack name is different.

**Release note**:

```other operator

```


ialidzhikov (commented):
Generally lgtm, 2 minor nits inline.
It also needs second opinion as I cannot say that I am very familiar with the existing flow and the recent change that caused the issue that is being fixed with this PR.

shaoyongfeng (approved):

Issue #202, opened 2020-12-14
cloud provider configuration will be deprecated for kubelet prameters


jia-jerry:
**How to categorize this issue?**

/area control-plane
/kind enhancement
/priority normal
/platform alicloud

**What would you like to be added**:
We need to move `provider-id`, `cloud-provider` and `enable-controller-attach-detach` from parameter to configuration file for kubelet
**Why is this needed**:
```
Dec 13 21:32:32 iZgw87uepwv66jorfgr1cmZ kubelet[2841]: Flag --provider-id has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
Dec 13 21:32:32 iZgw87uepwv66jorfgr1cmZ kubelet[2841]: Flag --cloud-provider has been deprecated, will be removed in 1.23, in favor of removing cloud provider code from Kubelet.
Dec 13 21:32:32 iZgw87uepwv66jorfgr1cmZ kubelet[2841]: Flag --enable-controller-attach-detach has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.```

rfranzke:
The `--cloud-provider` and `--enable-controller-attach-detach` will no longer be used for Shoots >= 1.23 as part of https://github.com/gardener/gardener-extension-provider-alicloud/pull/426.

However, I was not able to eliminate the `--provider-id` flag due to some special handling that is going on. There is already a TODO statement: https://github.com/gardener/gardener-extension-provider-alicloud/blob/c5bc4e7aceabe98ab8fdf211ea6682760ed8feac/pkg/webhook/controlplane/ensurer.go#L130-L135

It probably makes sense to follow up on this with the Alicloud Kubernetes community as I have no idea how to migrate this to the kubelet's configuration field.

/unassign

@jia-jerry @EmoinLanyu Please follow up with this.

shaoyongfeng:
With regards to provider-id flag, I did some investigation
1. providerid in node spec will be set by kubelet when initnode and joining cluster, if --provider-id flag is set as kubelet command line parameter
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L374-L376
2. how it works in cloud provider aws?
[...]

PR #426, opened 2022-01-26
Support for Kubernetes v1.23

rfranzke:

/area open-source usability
/kind enhancement
/platform alicloud
/exp intermediate
/topology garden seed shoot
/merge squash

This PR adds support for Kubernetes 1.23 to the extension.

Part of gardener/gardener#5102
Part of https://github.com/gardener/gardener-extension-provider-alicloud/issues/202

* I have successfully validated the functionality as follows:
  * ✅ Create new clusters with versions < 1.23
  * ✅ Create new clusters with version  = 1.23
  * ✅ Upgrade old clusters from version 1.22 to version 1.23
  * ✅ Delete clusters with versions < 1.23
  * ✅ Delete clusters with version  = 1.23

**Release note**:

```feature user
The Alicloud extension does now support shoot clusters with Kubernetes version 1.23. You should consider the [Kubernetes release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md) before upgrading to 1.23. 
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -23,11 +23,13 @@ import (
 	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
 	gcontext "github.com/gardener/gardener/extensions/pkg/webhook/context"
 	"github.com/gardener/gardener/extensions/pkg/webhook/controlplane/genericmutator"
-	versionutils "github.com/gardener/gardener/pkg/utils/version"
+	"github.com/gardener/gardener/pkg/utils/version"
+
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	"k8s.io/utils/pointer"
 )
 
 // NewEnsurer creates a new controlplane ensurer.
@@ -85,7 +87,7 @@ func ensureKubeAPIServerCommandLineArgs(c *corev1.Container, ver *semver.Version
 	c.Command = extensionswebhook.EnsureNoStringWithPrefixContains(c.Command, "--feature-gates=",
 		"CSIDriverRegistry=false", ",")
 
-	if versionutils.ConstraintK8sLess116.Check(ver) {
+	if version.ConstraintK8sLess116.Check(ver) {
 		c.Command = extensionswebhook.EnsureStringWithPrefixContains(c.Command, "--feature-gates=",
 			"ExpandCSIVolumes=true", ",")
 		c.Command = extensionswebhook.EnsureStringWithPrefixContains(c.Command, "--feature-gates=",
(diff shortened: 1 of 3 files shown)


rfranzke:
@EmoinLanyu Rebased

pkg/webhook/controlplane/ensurer.go
```diff
@@ -125,27 +127,35 @@ func (e *ensurer) EnsureKubeletServiceUnitOptions(ctx context.Context, gctx gcon
...
 	// offers a new `providerID` field which can be used, and it's expected that `--provider-id` will be deprecated eventually).
 	// Today, the problem is that the provider ID is determined dynamically using the script above, but the kubelet config cannot
 	// reference environment variables like it's possible today with the CLI parameters.
 	// See https://github.com/kubernetes/kubernetes/pull/90494
 	command = extensionswebhook.EnsureStringWithPrefix(command, "--provider-id=", "${PROVIDER_ID}")
-	command = extensionswebhook.EnsureStringWithPrefix(command, "--cloud-provider=", "external")
-	command = extensionswebhook.EnsureStringWithPrefix(command, "--enable-controller-attach-detach=", "true")
+
+	if !version.ConstraintK8sGreaterEqual123.Check(kubeletVersion) {
+		command = extensionswebhook.EnsureStringWithPrefix(command, "--cloud-provider=", "external")
```
shaoyongfeng, line 139: Hi @rfranzke ,
is "--cloud-provider" argument not needed at all? 
It is still kept in aws extension even if k8s version is older than 1.23
https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/webhook/controlplane/ensurer.go#L425

rfranzke: > is "--cloud-provider" argument not needed at all?

In my tests, I didn't see any issue when not providing it, and since it's deprecated with 1.23 I decided to longer provide it. Do you see any issues?

> It is still kept in aws extension even if k8s version is older than 1.23
https://github.com/gardener/gardener-extension-provider-aws/blob/master/pkg/webhook/controlplane/ensurer.go#L425

No, actually only when CSI is disabled which is only the case for < 1.18 AWS clusters, or when CSI is enabled but version is < 1.23

EmoinLanyu (approved):
/lgtm

pkg/webhook/controlplane/ensurer.go
```diff
@@ -125,27 +127,35 @@ func (e *ensurer) EnsureKubeletServiceUnitOptions(ctx context.Context, gctx gcon
...
 }
 
-func ensureKubeletCommandLineArgs(command []string) []string {
+func ensureKubeletCommandLineArgs(command []string, kubeletVersion *semver.Version) []string {
 	// TODO: Figure out how to provide the provider-id via the kubelet config file (as of Kubernetes 1.19 the kubelet config
 	// offers a new `providerID` field which can be used, and it's expected that `--provider-id` will be deprecated eventually).
 	// Today, the problem is that the provider ID is determined dynamically using the script above, but the kubelet config cannot
 	// reference environment variables like it's possible today with the CLI parameters.
 	// See https://github.com/kubernetes/kubernetes/pull/90494
 	command = extensionswebhook.EnsureStringWithPrefix(command, "--provider-id=", "${PROVIDER_ID}")
```
jia-jerry, line 136: Do we need to move `provider-id` to config file too?

rfranzke: Eventually yes, but not part of this PR. I have talked about this in https://github.com/gardener/gardener-extension-provider-alicloud/issues/202#issuecomment-1022443917

EmoinLanyu: > Do we need to move `provider-id` to config file too?

It is now addressed in TODO. For now I think we can keep it this way and figure out a solution later.

pkg/webhook/controlplane/ensurer.go
shaoyongfeng: ok.  Then it's safe to remove it

PR #436, opened 2022-02-17
inject kube token to alicloud diskplugin in the controller deployment

jia-jerry:

/area control-plane
/kind bug
/platform alicloud

AliCloud CSI need kubeconfig to create k8s go-client, though no k8s authorization are needed to be granted.
Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-alicloud/templates/auth/csi-attacher-rbac.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-alicloud/templates/auth/csi-attacher-rbac.yaml
@@ -16,7 +16,7 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["persistentvolumes"]
-  verbs: ["get", "list", "watch", "update"]
+  verbs: ["get", "list", "watch", "update", "patch"]
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["get", "list", "watch", "update"]
@@ -25,7 +25,7 @@ rules:
   verbs: ["get", "list", "watch"]
 - apiGroups: ["storage.k8s.io"]
   resources: ["volumeattachments"]
-  verbs: ["get", "list", "watch", "update"]
+  verbs: ["get", "list", "watch", "update", "patch"]
 {{- if semverCompare ">= 1.17" .Values.kubernetesVersion }}
 - apiGroups: [ "storage.k8s.io" ]
   resources: [ "volumeattachments/status" ]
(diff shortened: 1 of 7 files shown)


charts/internal/shoot-system-components/charts/csi-alicloud/templates/crd-volumesnapshotclasses.yaml
```diff
@@ -1,80 +0,0 @@
-{{- if semverCompare ">= 1.17" .Values.kubernetesVersion }}
```
ialidzhikov, line 1: Why you decided to remove these CRDs?

ialidzhikov (changes requested):

charts/internal/shoot-system-components/charts/csi-alicloud/templates/crd-volumesnapshotclasses.yaml
jia-jerry: I supposed they were not used. See [here](https://github.com/gardener/gardener-extension-provider-alicloud/blob/master/pkg/controller/controlplane/valuesprovider.go#L217). Did I make some mis-understanding?

charts/internal/shoot-system-components/charts/csi-alicloud/templates/auth/csi-attacher-rbac.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["persistentvolumes"]
-  verbs: ["get", "list", "watch", "update"]
+  verbs: ["get", "list", "watch", "update", "patch"]
```
jia-jerry, line 19: It seems wired those `patch` verbs are needed when projected token feature is enabled. 
The verbs should be required according to [csi-attacher](https://github.com/kubernetes-csi/external-attacher/blob/v3.3.0/pkg/controller/csi_handler.go#L786). But it worked before without this verb.
The same behavior occured to volumeattachments. It is interesting. @rfranzke  @ialidzhikov

charts/internal/shoot-system-components/charts/csi-alicloud/templates/crd-volumesnapshotclasses.yaml
jia-jerry: OK. BTW, how long do you plan to keep the legacy one?

jia-jerry:
Till now, this PR fixed all the CSI issue when projected-token-mount is used. Would you please review it?

rfranzke (approved):
Well done, thanks!

/lgtm

PR #449, opened 2022-03-03
Ensure resource cleanup and add integration test for invalid credentials

EmoinLanyu:

/area usability
/kind test
/platform alicloud

This PR tests https://github.com/gardener/terraformer/pull/93 and https://github.com/gardener/gardener/pull/3950 by adding an integration test case for invalid credentials.
It creates an infrastructure resource but uses invalid credentials and ensures that the ERR_INFRA_UNAUTHENTICATED is correctly detected from the termination message and that the infrastructure can still be deleted afterwards (as it's a no-op).

Before this PR, the destroy logic and testing was executed as part of a cleanup action, while the destroy logic should be part of the actual test and not just cleanup. Also, this could lead to resource leakage on AliCloud and on the test-machinery cluster, because if one cleanup action fails (e.g. the destroy logic) no further cleanup actions are executed.
Now, the destroy logic is executed as a deferred function to ensure that the cleanup actions are still run, even if the test itself fails.

Fixes #

Ref:
- https://github.com/gardener/gardener-extension-provider-aws/issues/436
- https://github.com/gardener/gardener-extension-provider-aws/pull/325
- https://github.com/gardener/gardener/pull/3950

**Release note**:

```other operator

```


jia-jerry:
please run `make revendor`

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -149,61 +151,143 @@ var _ = Describe("Infrastructure tests", func() {
 	})
 
 	Context("with infrastructure that requests existing vpc", func() {
-		var (
-			identifiers infrastructureIdentifiers
-		)
+		It("should successfully create and delete", func() {
+			identifiers := prepareVPC(ctx, clientFactory, *region, vpcCIDR, natGatewayCIDR)
+			defer func() {
+				cleanupVPC(ctx, clientFactory, identifiers)
```
jia-jerry, line 157: It will be better to put in clean up action. see here. https://github.com/gardener/gardener-extension-provider-aws/blob/master/test/integration/infrastructure/infrastructure_test.go#L205

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -149,61 +151,143 @@ var _ = Describe("Infrastructure tests", func() {
 	})
 
 	Context("with infrastructure that requests existing vpc", func() {
-		var (
-			identifiers infrastructureIdentifiers
-		)
+		It("should successfully create and delete", func() {
+			identifiers := prepareVPC(ctx, clientFactory, *region, vpcCIDR, natGatewayCIDR)
```
jia-jerry, line 155: we could simply puth vpc as a suite scpe resouce. so  you may put it in BeforeSuite.

test/integration/infrastructure/infrastructure_test.go
jia-jerry, line 157: https://github.com/gardener/gardener-extension-provider-aws/pull/325/files#r622763433. put it in `AddCleanupAction`

test/integration/infrastructure/infrastructure_test.go
jia-jerry, line 155: Would you please add a name for this VPC when creating it?

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -149,61 +150,143 @@ var _ = Describe("Infrastructure tests", func() {
...
-		AfterEach(func() {
-			framework.RunCleanupActions()
-			cleanupVPC(ctx, clientFactory, identifiers)
+			err := runTest(ctx, logger, c, providerConfig, decoder, clientFactory)
+			Expect(err).NotTo(HaveOccurred())
 		})
+	})
 
-		It("should successfully create and delete", func() {
+	Context("with invalid credentials", func() {
```
shaoyongfeng, line 169: I don't understand why we need this new test case. It doesn't improve residence of existing infrastructure testing

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -149,61 +151,143 @@ var _ = Describe("Infrastructure tests", func() {
...
 
-	var cleanupHandle framework.CleanupActionHandle
-	cleanupHandle = framework.AddCleanupAction(func() {
+	framework.AddCleanupAction(func() {
+		By("cleaning up namespace and cluster")
+		Expect(client.IgnoreNotFound(c.Delete(ctx, namespace))).To(Succeed())
+		Expect(client.IgnoreNotFound(c.Delete(ctx, cluster))).To(Succeed())
+	})
+
+	defer func() {
```
shaoyongfeng, line 272: If secret is rotated and not valid anymore, infra won't be deleted.  Then it's still not possible to remove namespace even you move the cleanup to aftereachsuite. Am I right?

test/integration/infrastructure/infrastructure_test.go
EmoinLanyu: As described in the PR description:
> This PR tests https://github.com/gardener/terraformer/pull/93 and https://github.com/gardener/gardener/pull/3950 by adding an integration test case for invalid credentials.
It creates an infrastructure resource but uses invalid credentials and ensures that the ERR_INFRA_UNAUTHENTICATED is correctly detected from the termination message and that the infrastructure can still be deleted afterwards (as it's a no-op).

This is added to capture another case that might happen during reconciliation.

test/integration/infrastructure/infrastructure_test.go
EmoinLanyu: Yes. If secret becomes invalid during the test, then the test here will not succeed in any case. Should that happen, currently we can only manually update the credentials in the TM cluster.

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -127,17 +131,15 @@ var _ = AfterSuite(func() {
 		mgrCancel()
 	}()
 
+	By("running cleanup actions")
+	framework.RunCleanupActions()
```
shaoyongfeng, line 135: Does it mean all namespaces generated during testing are only be cleaned at the end of testing?  Is it expected behavior?

test/integration/infrastructure/infrastructure_test.go
```diff
@@ -535,6 +624,7 @@ func prepareVPC(ctx context.Context, clientFactory alicloudclient.ClientFactory,
 	vpcClient, err := clientFactory.NewVPCClient(region, *accessKeyID, *accessKeySecret)
 	Expect(err).NotTo(HaveOccurred())
 	createVpcReq := vpc.CreateCreateVpcRequest()
+	createVpcReq.VpcName = "provider-alicloud-infra-test"
```
shaoyongfeng, line 627: what will happen if there are several testing running in parallel? will this name cause any conflict?

EmoinLanyu:
> will this name cause any conflict?

No it won't. VPC name is not required to be unique. So there is no problem here.

EmoinLanyu:
> Does it mean all namespaces generated during testing are only be cleaned at the end of testing? Is it expected behavior?

Yes the clean up actions are carried out at the end of the testing. The clean up actions will still be done should any tests fail.

shaoyongfeng (approved):

PR #451, opened 2022-03-07
support DISK_AD_CONTROLLER env for alicloud csi disk plugin daemonset

jia-jerry:

/area control-plane
/kind bug
/platform alicloud

support DISK_AD_CONTROLLER env for alicloud csi disk plugin daemonset

Fixes # 

**Release note**:

```feature user
User can specify DISK_AD_CONTROLLER environment for AliCloud CSI plugin via shoot manifest. It may Fix occasional volume attachment issue. By default, DISK_AD_CONTROLLER is true for newly created shoot.
```
```feature operator
csi block is introduced in ControlPlaneConfig in shoot. It may cause incompatible issue when the extension is rollback to previous version.
```


shaoyongfeng (dismissed):

shaoyongfeng (dismissed):

EmoinLanyu (approved):
/lgtm

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -516,12 +526,13 @@ func (vp *valuesProvider) getControlPlaneChartValues(
 	return values, nil
 }
 
-func (vp *valuesProvider) enableCSIADController() bool {
-	return vp.csi.EnableADController == nil || *vp.csi.EnableADController
+func (vp *valuesProvider) enableCSIADController(cpConfig *apisalicloud.ControlPlaneConfig) bool {
+	return vp.csi.EnableADController != nil && *vp.csi.EnableADController || cpConfig.CSI != nil && cpConfig.CSI.EnableADController != nil && *cpConfig.CSI.EnableADController
```
EmoinLanyu, line 530: So we should adjust the controller configuration and set `EnableADController` there to `false`. And let user decide whether they want this enabled for their old clusters.

jia-jerry: By default,  it is `false` if EnableADController  is not set in controller configuration.

EmoinLanyu: Yes, by default it is `false`. Just wondering if there is anywhere where we manually set it to `true`.

shaoyongfeng (approved):

PR #478, opened 2022-05-10
Removed validation blocking sfz

himanshu-kun:

/area auto-scaling
/kind bug
/platform alicloud

Removed a validation which stopped worker grp from having min 0.
Fixes #

A patch release would be needed as the other changes related to same feature are already out in the last release.

**Release note**:

```bugfix user
Users could configure worker pools to have min 0
```


shaoyongfeng:
Hi @himanshu-kun ，
pls cherry pick this pr to released branch if necessary

PR #488, opened 2022-05-26
admission-alicloud: Adapt Secrets webhook to rely on the provider label

ialidzhikov:
/area robustness
/area cost
/kind enhancement
/platform alicloud

This PR:
- Adapts the Secrets webhook to rely on the provider label (similar to https://github.com/gardener/gardener-extension-provider-gcp/pull/396). With this the Secrets webhook:
  1) no longer intercepts every Secret UPDATE request but only requests for Secrets with label `provider.shoot.gardener.cloud/alicloud="true"`
  2) no longer needs to list Shoots (hence, no longer has a cache for Shoots based on Informers) to understand whether a Secret is in-use by a Shoot from the same provider type ("alicloud")
- Adds a new SecretBinding webhook (similar to https://github.com/gardener/gardener-extension-provider-gcp/pull/428). This is to prevent creating SecretBindings that refer Secret from wrong provider type and hence block update request for this Secret as the admission will fail because the Secret does not contain valid credentials.

Part of https://github.com/gardener/gardener-extension-provider-gcp/issues/143 but for admission-alicloud

Unlike admission-{gcp,aws,azure} for admission-alicloud we cannot drop the cache for SecretBindings as the Shoot validator has to get the SecretBinding -> then the Secret and call the Alicloud API to get the nat gateway zones:
https://github.com/gardener/gardener-extension-provider-alicloud/blob/b823ea14db3e132b6ae7500321a6ebe3e9553f7d/pkg/admission/validator/shoot.go#L214-L246

This is actually problematic because if the request to alicloud APIs fail because of some reason then the admission-alicloud will reject any create/update request for Shoots. -> https://github.com/gardener/gardener-extension-provider-alicloud/issues/489

**Release note**:

```breaking operator
This version of admission-alicloud requires the SecretBinding provider controller to be enabled - enabled by default for gardener-controller-manager >= 1.42 or can be enabled via the gardener-controller-manager component config.
```

```other operator
The Secrets webhook of admission-alicloud:
- no longer intercepts every Secret UPDATE request but only requests for Secrets that are associated with a SecretBinding with `provider.type=alicloud`.
- no longer needs to list Shoots (hence, no cache for Shoots)
```

```other operator
The admission-alicloud component introduces a new SecretBinding validator. It validates requests for SecretBindings and checks whether the SecretBinding refers to a valid Alicloud Secret.
```


pkg/admission/validator/secretbinding.go
```diff
@@ -0,0 +1,76 @@
...
+		oldSecretBinding, ok := oldObj.(*core.SecretBinding)
+		if !ok {
+			return fmt.Errorf("wrong object type %T for old object", oldObj)
+		}
+
+		// If the provider type did not change, we exit early.
+		if oldSecretBinding.Provider != nil && equality.Semantic.DeepEqual(secretBinding.Provider.Type, oldSecretBinding.Provider.Type) {
+			return nil
+		}
+	}
```
EmoinLanyu, line 63: Makes sense since no other field is mutable. On a side note, test case for update is missing.

EmoinLanyu, line 63: Makes sense since no other field is mutable. On a side note, test case for update is missing.

EmoinLanyu (approved):

PR #503, opened 2022-06-22
Add parsers for csi-snapshot pods

Kristian-ZH:

/area logging
/kind enhancement
/platform alicloud

Add parsers for csi-snapshot pods

Fixes #

**Release note**:

```other operator
None
```


charts/internal/seed-controlplane/charts/csi-alicloud/templates/csi-snapshot-controller.yaml
```diff
@@ -28,7 +28,7 @@ spec:
     spec:
       automountServiceAccountToken: false
       containers:
-      - name: csi-snapshot-controller
+      - name: alicloud-csi-snapshot-controller
```
shaoyongfeng, line 31: if you change container name, you may need to go through all places where uses it and change accordingly
https://github.com/gardener/gardener-extension-provider-alicloud/search?q=csi-snapshot-validation

Kristian-ZH: So, the only place I have to change it is in the [charts/internal/seed-controlplane/charts/csi-alicloud/templates/csi-snapshot-validation-webhook-vpa.yaml](https://github.com/gardener/gardener-extension-provider-alicloud/blob/30d72a0273a1af2a9d44eecdbd83d603635e2832/charts/internal/seed-controlplane/charts/csi-alicloud/templates/csi-snapshot-validation-webhook-vpa.yaml), correct?

All other places you linked does not link the container

shaoyongfeng (approved):

PR #517, opened 2022-07-26
allow all communication from pod ip range

DockToFuture:

/area networking
/kind enhancement
/platform alicloud

To get rid of the networking overlay we need to allow the pod ip range communication within the vpc network.
Prerequisite for https://github.com/gardener/backlog/issues/29.

Fixes #

**Release note**:

```other operator
none
```

### DIFF
--- a/pkg/controller/infrastructure/templates/main.tpl.tf
+++ b/pkg/controller/infrastructure/templates/main.tpl.tf
@@ -122,6 +122,16 @@ resource "alicloud_security_group_rule" "allow_all_internal_udp_in" {
   cidr_ip           = "{{ .vpc.cidr }}"
 }
 
+resource "alicloud_security_group_rule" "allow_all_internal_pod_traffic_in" {
+  type              = "ingress"
+  ip_protocol       = "all"
+  policy            = "accept"
+  port_range        = "-1/-1"
+  priority          = 1
+  security_group_id = alicloud_security_group.sg.id
+  cidr_ip           = "{{ .podCIDR }}"
+}
+
 // We have introduced new output variables. However, they are not applied for
 // existing clusters as Terraform won't detect a diff when we run `terraform plan`.
 // Workaround: Providing a null-resource for letting Terraform think that there are
--- a/pkg/controller/infrastructure/actuator.go
+++ b/pkg/controller/infrastructure/actuator.go
@@ -196,8 +196,8 @@ func (a *actuator) getInitializerValues(
 	return a.terraformChartOps.ComputeUseVPCInitializerValues(config, vpcInfo), nil
 }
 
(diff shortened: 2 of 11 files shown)


ScheererJ (dismissed):
/lgtm

shaoyongfeng:
HI @DockToFuture ，
Integration test fails with error. can you check it?


2022-08-05 11:20:59 | 2022-08-05T03:20:59.729157677Z stdout F     1.659668733085154e+09	ERROR	Reconciler error	{"controller": "infrastructure_controller", "object": {"name":"infrastructure","namespace":"provider-alicloud-test-h9vlb"}, "namespace": "provider-alicloud-test-h9vlb", "name": "infrastructure", "reconcileID": "4491569a-bd22-4e06-b719-90cae0716144", "error": "panic: runtime error: invalid memory address or nil pointer dereference [recovered]"}
-- | --

ScheererJ:
Updated end-to-end test accordingly so that it should go through now.

pkg/controller/infrastructure/templates/main.tpl.tf
```diff
@@ -122,6 +122,16 @@ resource "alicloud_security_group_rule" "allow_all_internal_udp_in" {
...
+resource "alicloud_security_group_rule" "allow_all_internal_pod_traffic_in" {
+  type              = "ingress"
+  ip_protocol       = "all"
+  policy            = "accept"
+  port_range        = "-1/-1"
+  priority          = 1
+  security_group_id = alicloud_security_group.sg.id
+  cidr_ip           = "{{ .podCIDR }}"
+}
+
```
shaoyongfeng, line 134: Hi,
this security group rule will be created no matter user uses overlay network or not.  is there any harm?

is there a way for user to disable overlay network for existing shoot cluster?  If yes, we also need to consider how to trigger infrastructure reconlication in order to get this security rule created.

ScheererJ: Hi,

this security group rule will allow packets in the pod ip range to flow through the virtual network. In case an overlay network is used this rule allows packets, which will never be created. Hence, it is not required in this case, but it also does not harm as on the kubernetes nodes there are anyway local routing rules in place preventing packets in the pod ip range to go out directly if an overlay network is used.
So in short: From my perspective, there is no harm.

It was possible all the time to disable the overlay network in the configuration of the networking extension. However, this would have led to failing cluster in the past. Hence, I do not see an urgent need to add a reconcile flow now. Switching the overlay network is an incompatible operation causing at least some downtime in terms of network connectivity. We are not yet at the stage that we can support this. Therefore, I would not necessarily do it for old clusters, but only for new ones.

shaoyongfeng (approved):

PR #518, opened 2022-07-29
run new shoot clusters without overlay if not explicitly specified

DockToFuture:

/area networking
/kind enhancement
/platform alicloud

A mutating webhook is added, so that all new calico shoot clusters on alicloud will be created without an overlay if not explicitly specified in the shoot spec.
Prerequisite for https://github.com/gardener/backlog/issues/29.

Fixes #

**Release note**:

```other operator
All new calico alicloud shoot clusters will be created without an overlay if not explicitly specified in the shoot spec.
```


ScheererJ:
Rebased as requested.

pkg/webhook/network/mutate.go
```diff
@@ -0,0 +1,75 @@
...
+
+	calicov1alpha1 "github.com/gardener/gardener-extension-networking-calico/pkg/apis/calico/v1alpha1"
+	calicov1alpha1helper "github.com/gardener/gardener-extension-networking-calico/pkg/apis/calico/v1alpha1/helper"
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
+)
+
+func mutateNetworkConfig(new, old *extensionsv1alpha1.Network) error {
```
shaoyongfeng, line 28: what will happen if shoot uses cilium network instead of calico? will networkconfig also be mutated?

shaoyongfeng (approved):

PR #533, opened 2022-09-14
just check whether snat entry is empty when checking snattable

shaoyongfeng:
/area testing
/kind bug
/platform alicloud

Ali changes the behavior when snatable doesn't exist.  When creating infra with new VPC, snat table is deleted when destroying infra. In this case, an error is returned by alicloud
Fixes #

**Release note**:

```other operator
none
```


EmoinLanyu (approved):
/lgtm

PR #640, opened 2023-08-22
change default disktype from ssd to essd

shaoyongfeng:
/kind enhancement 
/platform alicloud

Switch default storage class to use `cloud_essd` since AliCloud is deprecating cloud_ssd. This will affect only newly created disk.

**Release note**:
```
 The default storage class is now using `cloud_essd` disk type instead of `cloud_ssd` . This will affect only newly created disks.
```


dguendisch (approved):
/lgtm

kevin-lacoo (approved):
LGTM

PR #718, opened 2024-05-06
switch to docker-buildx as oci-builder

ccwienk:
Support for kaniko is planned to be dropped in pipeline-template. Switch to docker-buildx to unblock upstream removal of kaniko-code. This change should be a drop-in-replacement with no additional required actions.

/area delivery
/kind cleanup
/platform alicloud

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -6,7 +6,7 @@ gardener-extension-provider-alicloud:
       version:
         preprocess: 'inject-commit-hash'
       publish:
-        oci-builder: 'kaniko'
+        oci-builder: 'docker-buildx'
         dockerimages:
           gardener-extension-provider-alicloud:
             registry: 'gcr-readwrite'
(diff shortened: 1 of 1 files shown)


shaoyongfeng:
Hi @ccwienk ,
Is it intended to have a PR to [release-v1.50-ccm-metrics-port](https://github.com/gardener/gardener-extension-provider-alicloud/tree/release-v1.50-ccm-metrics-port) branch

ccwienk:
@shaoyongfeng : yes. I have been creating a couple of pull-requests for stale branches (where last commit was older than ~200d) to both free resources and also get rid of pipelines still using `kaniko` oci builder.

After that cleanup, there were a few pipelines remaining, which were not stale, but still using kaniko builder. For those, I subsequently created pullrequests like this one, which retains configured pipelines, but switches OCI Builder to `docker buildx`. The pipeline will remain fully functional after accepting this change.

shaoyongfeng (approved):

PR #765, opened 2025-02-14
Forbid privilege escalations for containers

georgibaltiev:

/area security
/area compliance
/kind enhancement
/platform alicloud

This PR sets the `securityContext.allowPrivilegeEscalation` field to `false` for every container, which does not have `securityContext.Privileged` set to `true` or one of `CAP_SYS_ADMIN/SYS_ADMIN` capabilities added.

Part of gardener/gardener#11139

cc @AleksandarSavchev 

**Release note**:

```other operator
Containers, which do not require privilege escalations, now forbid privilege escalations explicitly.
```


charts/internal/shoot-system-components/charts/csi-alicloud/templates/csi-diskplugin-ds.yaml
```diff
@@ -102,6 +104,8 @@ spec:
         resources:
 {{ toYaml .Values.resources.driver | indent 10 }}
 {{- end }}
+        securityContext:
+          allowPrivilegeEscalation: false
```
AleksandarSavchev, line 108:
```suggestion
```
We should remove this as the container has `securityContext.privileged: true`

AleksandarSavchev (changes requested):
Thanks, only one minor change needed.

kevin-lacoo (approved):

PR #787, opened 2025-05-13
ReadSecretCredentials: Only return credentialsFile when present

AleksandarSavchev:

/area quality
/kind bug
/platform alicloud

The `ReadSecretCredentials()` function was changed to only return `CredentialsFile` when present in the specified secret. The field is not required in all cases and could cause failures in `BackupBucket` reconciliation [ref](https://github.com/gardener/gardener-extension-provider-alicloud/blob/678c5e99bc68fef449bd9eaba9865ef619088cf5/pkg/controller/backupbucket/actuator.go#L33)

Fixes #

**Release note**:

```bugfix user
A bug was fixed which caused `BackupBucket`s to fail in reconciliation when the referenced secret did not contain a `credentialsFile` field.
```


ialidzhikov (approved):
/lgtm

shaoyongfeng (approved):

PR #873, opened 2026-01-29
[GEP-33] Add machineCapabilities CloudProfile support

kevin-lacoo:

/area control-plane
/kind enhancement
/platform alicloud

This PR is part of [GEP-33 Machine Image Capabilities](https://github.com/gardener/gardener/blob/master/docs/proposals/33-machine-image-capabilities.md)
Part of: https://github.com/gardener/gardener/issues/11301

**Release note**:

```other operator
Adopts Gardener MachineImage `Capabilities` and introduces `CapabilityFlavors` to the `providerConfig`.
```


pkg/apis/alicloud/types_cloudprofile.go
```diff
@@ -31,8 +32,24 @@ type MachineImages struct {
...
+// MachineImageFlavor groups all RegionIDMappings for a specific set of capabilities.
+type MachineImageFlavor struct {
+	// Regions is a mapping to the correct ID for the machine image in the supported regions.
+	Regions []RegionIDMapping
+	// Capabilities that are supported by the machine image IDs in this set.
+	Capabilities gardencorev1beta1.Capabilities
+}
+
+// GetCapabilities returns the Capabilities of a MachineImageFlavor
+func (cs *MachineImageFlavor) GetCapabilities() gardencorev1beta1.Capabilities {
```
kevin-lacoo:
```suggestion
func (cs MachineImageFlavor) GetCapabilities() gardencorev1beta1.Capabilities {
```

pkg/apis/alicloud/v1alpha1/types_cloudprofile.go
```diff
@@ -32,8 +33,24 @@ type MachineImages struct {
...
+// MachineImageFlavor groups all RegionIDMappings for a specific set of capabilities.
+type MachineImageFlavor struct {
+	// Regions is a mapping to the correct ID for the machine image in the supported regions.
+	Regions []RegionIDMapping `json:"regions"`
+	// Capabilities that are supported by the machine image IDs in this set.
+	Capabilities gardencorev1beta1.Capabilities `json:"capabilities,omitempty"`
+}
+
+// GetCapabilities returns the Capabilities of a MachineImageFlavor
+func (cs *MachineImageFlavor) GetCapabilities() gardencorev1beta1.Capabilities {
```
kevin-lacoo:
```suggestion
func (cs MachineImageFlavor) GetCapabilities() gardencorev1beta1.Capabilities {
```

pkg/controller/common/encryptimage.go
```diff
@@ -233,11 +233,11 @@ func (ie *imageEncryptor) getEncrytpedImageIDFromStack(stackId string) (string,
...
 func (ie *imageEncryptor) getStackName() string {
-	return GetEncryptImageStackName(ie.imageName, ie.imageVersion, ie.regionID)
+	return GetEncryptImageStackName(ie.imageName, ie.imageVersion, ie.regionID, ie.sourceImageID)
 }
 
 // GetEncryptImageStackName returns the encrypt image stack name for the given image name and version.
-func GetEncryptImageStackName(imageName, imageVersion, regionID string) string {
-	var rosNameFormat = "encrypt_image_%s_%s_%s"
-	return strings.ReplaceAll(fmt.Sprintf(rosNameFormat, imageName, imageVersion, regionID), ".", "-")
+func GetEncryptImageStackName(imageName, imageVersion, regionID, imageId string) string {
```
Roncossek, line 240: This is a breaking change for existing encrypted image stacks, istn't it? When a shoot reconciles after this update, it will look for a stack with the new naming format and wont find the existing one, am i wrong?

pkg/apis/alicloud/helper/helper.go
```diff
@@ -111,3 +145,165 @@ func FindImageForRegionFromCloudProfile(cloudProfileConfig *api.CloudProfileConf
...
+			// image is already legacy format
+			return img
+		} else {
+			return &api.MachineImage{
+				Name:      img.Name,
+				Version:   img.Version,
+				ID:        img.ID,
+				Encrypted: img.Encrypted,
+			}
+		}
```
Roncossek: nit. else not require here.
```suggestion
			return &api.MachineImage{
				Name:      img.Name,
				Version:   img.Version,
				ID:        img.ID,
				Encrypted: img.Encrypted,
			}
```

pkg/apis/alicloud/helper/helper.go
```diff
@@ -111,3 +145,165 @@ func FindImageForRegionFromCloudProfile(cloudProfileConfig *api.CloudProfileConf
...
+		return img
+	} else {
+		return &api.MachineImage{
+			Name:         img.Name,
+			Version:      img.Version,
+			ID:           img.ID,
+			Encrypted:    img.Encrypted,
+			Capabilities: gardencorev1beta1.Capabilities{v1beta1constants.ArchitectureName: []string{v1beta1constants.ArchitectureAMD64}},
+		}
+	}
```
Roncossek: nit. else not require here.
```suggestion
		return &api.MachineImage{
			Name:         img.Name,
			Version:      img.Version,
			ID:           img.ID,
			Encrypted:    img.Encrypted,
			Capabilities: gardencorev1beta1.Capabilities{v1beta1constants.ArchitectureName: []string{v1beta1constants.ArchitectureAMD64}},
		}
```

test/integration/infrastructure/encryptedimagehelper.go
```diff
@@ -28,6 +28,7 @@ import (
 
 var imageName = "ubuntu"
 var imageVersion = "20.04"
+var imageId = getImageId(*region)
```
Roncossek, line 31: not related to this PR, but I guess this file name is off and should be ether `encryptedimagehelper.go` (with an added `d`) or `encryptimagehelper.go` (with removed `e`)

pkg/admission/validator/namespacedcloudprofile.go
```diff
@@ -146,11 +150,125 @@ func (p *namespacedCloudProfile) validateMachineImages(providerConfig *api.Cloud
...
+						machineImage.Name, version.Version, region, coreDefaultedCapabilityFlavor.Capabilities)))
+			}
+		}
+	}
+
+	return allErrs
+}
+
+// validateRegionsFormatWithCapabilities validates when provider uses old format (no architecture in regions all defalt to amd64) in capabilities CloudProfile
+func validateRegionsFormatWithCapabilities(machineImage core.MachineImage, version gardencorev1beta1.MachineImageVersion, _ api.MachineImageVersion, defaultedCapabilityFlavors []gardencorev1beta1.MachineImageFlavor, path *field.Path) field.ErrorList {
```
Roncossek: Why keep an unused parameter?
```suggestion
func validateRegionsFormatWithCapabilities(machineImage core.MachineImage, version gardencorev1beta1.MachineImageVersion, defaultedCapabilityFlavors []gardencorev1beta1.MachineImageFlavor, path *field.Path) field.ErrorList {
```

Roncossek:
https://github.com/gardener/gardener-extension-provider-alicloud/blob/81231eefa861ce76699c347383b828d0fb49f24c/pkg/controller/bastion/actuator_reconcile.go#L65
When only `architecture` with `amd64` is supported this no issue, but now with support for arbitrary capabilities there needs to be an additional check, that the capabilities of the first image in the infrastructure status are compatible with the machineType used to create the bastion.

Similar to whats happening for e.g. the aws-profider [here](https://github.com/gardener/gardener-extension-provider-aws/blob/17d43e5d20d304c285ce1bf06d2d3270118cc180/pkg/controller/bastion/options.go#L107)

pkg/apis/alicloud/helper/helper.go
kevin-lacoo: code updated

pkg/apis/alicloud/helper/helper.go
kevin-lacoo: code updated

pkg/admission/validator/namespacedcloudprofile.go
kevin-lacoo: code updated

test/integration/infrastructure/encryptedimagehelper.go
kevin-lacoo: the file name is changed.

pkg/controller/common/encryptimage.go
kevin-lacoo: Hello, the following is the explanation regarding the change of the stack name:
For existing shoots, when calling **ensureEncryptedImageForShootProviderAccount,** the first step is to check whether the required image (name, version, encryption) already exists in **infrastructureStatus.MachineImages**. If it exists, the subsequent steps of searching for the image based on capability, encrypting the image, and calling **GetEncryptImageStackName** will be skipped. [Here](https://github.com/kevin-lacoo/gardener-extension-provider-alicloud/blob/6f49285f0f58cf5ec88c8b4fcd1fe790a1c073c1/pkg/controller/infrastructure/shoot_encrypted_image.go#L118) is the code.
Only when the name, version, or encryption of the required image changes will the subsequent steps of searching for the image and encrypting it be executed. At this time, the change in the stack name is irrelevant. 
For multiple shoots under the same IAAS account, when creating a new shoot, it is indeed possible to re-encrypt previously encrypted images (due to the change in the stack name). However, this is part of the new shoot creation process, and even without using the new provider version, the image encryption will still be required. Subsequent new shoot creations will not require re-encryption, as the new stack name already exists.

kevin-lacoo:
> https://github.com/gardener/gardener-extension-provider-alicloud/blob/81231eefa861ce76699c347383b828d0fb49f24c/pkg/controller/bastion/actuator_reconcile.go#L65
> 
> 
> When only `architecture` with `amd64` is supported this no issue, but now with support for arbitrary capabilities there needs to be an additional check, that the capabilities of the first image in the infrastructure status are compatible with the machineType used to create the bastion.
> Similar to whats happening for e.g. the aws-profider [here](https://github.com/gardener/gardener-extension-provider-aws/blob/17d43e5d20d304c285ce1bf06d2d3270118cc180/pkg/controller/bastion/options.go#L107)

Because the native image of Alibaba is not encrypted, we need to encrypt the required image during the reconciliation of infrastructure and record the encrypted image in i**nfrastructureStatus.MachineImages** for subsequent modules to use. Therefore, bastion needs to retrieve the image from **infrastructureStatus.MachineImages**. 
I have extended the bastion code to support ARM-type images and machineType (originally only supporting X86). For other non-architecture capabilities, Alibaba currently does not have them, which can be further enhanced when needed in the future.

pkg/controller/common/encryptimage.go
Roncossek: Thanks for the clarification.

Roncossek:
> > https://github.com/gardener/gardener-extension-provider-alicloud/blob/81231eefa861ce76699c347383b828d0fb49f24c/pkg/controller/bastion/actuator_reconcile.go#L65
> > 
> > When only `architecture` with `amd64` is supported this no issue, but now with support for arbitrary capabilities there needs to be an additional check, that the capabilities of the first image in the infrastructure status are compatible with the machineType used to create the bastion.
> > Similar to whats happening for e.g. the aws-profider [here](https://github.com/gardener/gardener-extension-provider-aws/blob/17d43e5d20d304c285ce1bf06d2d3270118cc180/pkg/controller/bastion/options.go#L107)
> 
> Because the native image of Alibaba is not encrypted, we need to encrypt the required image during the reconciliation of infrastructure and record the encrypted image in i**nfrastructureStatus.MachineImages** for subsequent modules to use. Therefore, bastion needs to retrieve the image from **infrastructureStatus.MachineImages**. I have extended the bastion code to support ARM-type images and machineType (originally only supporting X86). For other non-architecture capabilities, Alibaba currently does not have them, which can be further enhanced when needed in the future.

interesting. But what about the next scenario: Even though this extension only knows about the `architecture`-capability. The operator can (in theory) already define its own custom capabilities (like `gpu-support` or `hypervisorType` .... ) these should be taken into consideration when chosing an image for a given machine. This happens for workers, but is missing for bastions. So this is an architectural gap.

shaoyongfeng:
> > > https://github.com/gardener/gardener-extension-provider-alicloud/blob/81231eefa861ce76699c347383b828d0fb49f24c/pkg/controller/bastion/actuator_reconcile.go#L65
> > > 
> > > When only `architecture` with `amd64` is supported this no issue, but now with support for arbitrary capabilities there needs to be an additional check, that the capabilities of the first image in the infrastructure status are compatible with the machineType used to create the bastion.
> > > Similar to whats happening for e.g. the aws-profider [here](https://github.com/gardener/gardener-extension-provider-aws/blob/17d43e5d20d304c285ce1bf06d2d3270118cc180/pkg/controller/bastion/options.go#L107)
> > 
> > 
> > Because the native image of Alibaba is not encrypted, we need to encrypt the required image during the reconciliation of infrastructure and record the encrypted image in i**nfrastructureStatus.MachineImages** for subsequent modules to use. Therefore, bastion needs to retrieve the image from **infrastructureStatus.MachineImages**. I have extended the bastion code to support ARM-type images and machineType (originally only supporting X86). For other non-architecture capabilities, Alibaba currently does not have them, which can be further enhanced when needed in the future.
> 
> interesting. But what about the next scenario: Even though this extension only knows about the `architecture`-capability. The operator can (in theory) already define its own custom capabilities (like `gpu-support` or `hypervisorType` .... ) these should be taken into consideration when chosing an image for a given machine. This happens for workers, but is missing for bastions. So this is an architectural gap.

Hi Johannes,
After internally discussion, we would like to leave as it is.  Bastions are only used during trouble shooting and not part of cluster.  Machine type used for bastion is not even from cloud profile but get a cheapest one via AliCloud API.  Therefore, it's even not possible to use the same logic to match images and machine type.

shaoyongfeng (approved):

PR #901, opened 2026-05-11
Enable custom route table and dual stack

kevin-lacoo:

/area networking
/kind enhancement
/platform alicloud

This PR introduce following two features to provider-alicloud:
***Custom Route Table (networks.vpc.useCustomRouteTable)***
When multiple shoots share a user-provided VPC, they all write pod-to-node routes into the VPC's system default route table by default. This causes route conflicts between shoots. With useCustomRouteTable: true, Gardener creates a dedicated route table per shoot and associates all shoot VSwitches to it. The CCM receives this route table ID via routeTableIDS and writes pod routes there instead of the system default table — eliminating the conflict.
```
spec:
  provider:
    infrastructureConfig:
      networks:
        vpc:
          useCustomRouteTable: true
```

***Dual-Stack Support (dualStack.enabled)***
Enables Dual-stack networking for the shoot. When set to true:
Once dual-stack is enabled, the Cloud Controller Manager can provision DualStack NLB service.
For Gardener-managed VPCs: Gardener upgrades the VPC to IPv6, creates an IPv6 Gateway, and assigns /64 IPv6 subnets to each VSwitch according to networks.zones[].ipv6CidrBlock (0–255).
For user-provided VPCs: The VPC must already have IPv6 and an IPv6 Gateway enabled by the user. Gardener assigns /64 subnets to VSwitches per ipv6CidrBlock.
A ::/0 → IPv6Gateway route is added to the relevant route table (system default or custom).
ipv6CidrBlock is required for every zone when dual-stack is enabled, values must be unique across zones.
```
spec:
  provider:
    infrastructureConfig:
      dualStack:
        enabled: true
```

Fixes #

**Release note**:

```other operator
Infrastructure custom route table support can be enabled via `spec.provider.infrastructureConfig.networks.vpc.useCustomRouteTable: true` in the shoot.yaml.
```
```other operator
Infrastructure dualstack support can be enabled via `spec.provider.infrastructureConfig.dualStack.enabled: true` in the shoot.yaml.
```


pkg/apis/alicloud/validation/infrastructure.go
```diff
@@ -5,16 +5,48 @@
...
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	apisalicloud "github.com/gardener/gardener-extension-provider-alicloud/pkg/apis/alicloud"
 )
 
+// nlbSupportedRegions lists Alicloud regions that support NLB (last updated: 2026-04).
+// Keep in sync with https://help.aliyun.com/zh/slb/network-load-balancer/product-overview/regions-that-support-nlb
+var nlbSupportedRegions = sets.New[string](
```
hebelsan, line 22: I’m not sure this hardcoded list is maintainable long term.
I’d rather fetch the data from the API instead.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -379,3 +451,147 @@ func (c *FlowContext) ensureManagedNatGateway(ctx context.Context) error {
...
+		DestinationCidrBlock: "0.0.0.0/0",
+		NextHopType:          "NatGateway",
+		NextHopId:            *natGwId,
+		Name:                 c.namespace + "-rt-default",
+	}); err != nil {
+		return err
+	}
+
+	// Ensure ::/0 → IPv6Gateway when dualStack is enabled
+	if c.dualStackEnabled() {
```
hebelsan, line 543: This code path is only reached if `useCustomRouteTable=true`.
Does Alicloud's platform automatically inserts a system ::/0 route into the VPC's default route table when an IPv6 Gateway is created?

hebelsan (commented):
Since I know very little about Alibaba Cloud, please take this review with a grain of salt.
Also, even when changes are related, please split them into separate PRs.
It makes the review process much easier.

pkg/apis/alicloud/validation/infrastructure.go
kevin-lacoo: Hi hebelsan, there are two reasons for not using API: 
1. This is a validation check on the region of shoot in admission, and to avoid API exceptions, we usually avoid calling API in admission validation. 
3. Alibaba does have an API to query regions supporting NLB, but not all regions supporting NLB support dual-stack, so it's not feasible to use API.

pkg/controller/infrastructure/infraflow/reconcile.go
kevin-lacoo: Yes, Alicloud's platform will automatically inserts a system ::/0 route into the VPC's default route table when an IPv6Gateway is created, and only one IPv6Gateway can be created in one VPC.

pkg/apis/alicloud/validation/infrastructure.go
hebelsan: okay, I understand.

shaoyongfeng (approved):

